PR #1, opened 2024-09-17
Incident regression in r

eschrom:
This is the first version of my incident uptake regression method for uptake projections. 
- R contains the functions
- scripts contains a short script using the R functions
- data contains .csv files with full names and abbreviations of geographic regions (state and country)

The primary aim is to import data into a standardized format, fit a regression model on incident uptake, and use that regression model to make projections (either genuine predictions of the future or predictions of the data the model was trained on). There are also functions for interpolating dates and scaling one data set to another.

@Fuhan-Yang please review to determine if this works for now (given the impending switch to Python), or if you'd like any changes. @swo please also feel free to review and/or comment. Thank you!

### DIFF
--- a/R/build_projection_model.r
+++ b/R/build_projection_model.r
@@ -1,0 +1,64 @@
+#' Model future uptake from current uptake
+#'
+#' BEWARE: THIS DOES NOT YET ACCOUNT FOR MULTI-YEAR OR MULTI-STATE DATA SETS
+#'
+#' Future incident uptake can be predicted from current incident uptake.
+#' A model can be built and fit to project incident uptake into the future.
+#'
+#' The default model says that current incident uptake is a linear function
+#' of the previous incident uptake, time elapsed since rollout, and their
+#' interaction.
+#'
+#' All modeling is performed on the daily-average incident uptake to
+#' remove accumulation dependence among data points and to
+#' account for slight variations in incident reporting intervals.
+#' The first data point may be dropped before fitting the model
+#' if there is a long lag between the vaccine rollout date and
+#' the date that data collection began.
+#' All variables are standardized prior to fitting the model.
+#'
+#' UPDATES FOR THE FUTURE:
+#' - Model must incorporate year-to-year variability
+#' - Model must incorporate state-to-state variability
+#' - A variety of models should be tried and competed, or formula provided
+#'
+#' @description Fit a model to project uptake into the future
+#'
+#' @param data uptake data to fit the projection model
+#' @param drop_first whether to drop the first data point
+#'
+#' @return a brms model object
+#'
+#' @export
+#'
+build_projection_model <- function(data, drop_first = NULL) {
+
+  if (is.null(drop_first)) {
+    intervals <- diff(data$elapsed)
+    if (abs(data$elapsed[1] - mean(intervals)) > (sd(intervals) + 1)) {
+      drop_first <- TRUE
+    } else {
+      drop_first <- FALSE
+      intervals <- c(data$elapsed[1], intervals)
+    }
+  }
+
+  if (drop_first) {
+    data <- data[-1, ]
+  }
+
+  data$previous <- c(NA, data$daily[-nrow(data)])
+  data <- data[-1, ]
+  data$previous_std <- (data$previous - mean(data$previous)) / sd(data$previous)
+  data$elapsed_std <- (data$elapsed - mean(data$elapsed)) / sd(data$elapsed)
+  data$daily_std <- (data$daily - mean(data$daily)) / sd(data$daily)
+
+  model <- brms::brm(daily_std ~ previous_std * elapsed_std,
+                     data = data,
+                     family = gaussian())
+
+  print(summary(model))
+  plot(model)
+
+  return(model)
+}
--- a/R/build_scaling_model.r
+++ b/R/build_scaling_model.r
@@ -1,0 +1,89 @@
+#' Model one source of uptake data from another
+#'
+#' Different sources of uptake data may systematically disagree.
+#' A model can be built and fit to scale one source to the other.
+#'
+#' The default model says that the uptake data from the target source
+#' is a linear function of the input data, time elapsed since rollout,
+#' and their interaction. If multiple states are represented, each state
+#' is modeled separately via group-level variation of the intercept and
+#' the input data coefficients.
+#'
+#' All modeling is performed on the daily-average incident uptake to
+#' remove accumulation dependence among data points and to
+#' account for slight variations in incident reporting intervals.
+#' The first data point may be dropped before fitting the model
+#' if there is a long lag between the vaccine rollout date and
+#' the date that data collection began.
+#' All variables are standardized prior to fitting the model.
+#'
+#' UPDATES FOR THE FUTURE:
+#' - Model must incorporate year-to-year variability
+#' - A variety of models should be tried and competed, or formula provided
+#'
+#' @description Fit a model to scale uptake date from one source to another
+#'
+#' @param input uptake data to use as a predictor
+#' @param target uptake data to use as the outcome
+#' @param drop_first whether to drop the first data point
+#'
+#' @return a brms model object
+#'
+#' @export
+#'
+build_scaling_model <- function(input, target, drop_first = NULL) {
+
+  if (is.null(drop_first)) {
+    input_intervals <- c(input$elapsed[1], diff(input$elapsed))
+    target_intervals <- c(target$elapsed[1], diff(target$elapsed))
+
+    common_dates <- intersect(input$elapsed, target$elapsed)
+
+    input_intervals <- input_intervals[input$elapsed %in% common_dates]
+    target_intervals <- target_intervals[target$elapsed %in% common_dates]
+
+    if ((input_intervals[1] - mean(input_intervals[-1])) >
+          (sd(input_intervals[-1]) + 1) ||
+          (target_intervals[1] - mean(target_intervals[-1])) >
+            (sd(target_intervals[-1]) + 1)) {
+      drop_first <- TRUE
+    } else {
+      drop_first <- FALSE
+    }
+  }
+
+  input <- dplyr::semi_join(input, target, by = "date")
+  target <- dplyr::semi_join(target, input, by = "date")
+
+  if (drop_first) {
+    input <- input[-1, ]
+    target <- target[-1, ]
+  }
+
+  if (nrow(input) != nrow(target)) {
+    stop("There are state or date mismatches between the input and target data")
+  }
+
+  input$daily_std <- (input$daily - mean(input$daily)) /
+    sd(input$daily)
+  input$elapsed_std <- (input$elapsed - mean(input$elapsed)) /
+    sd(input$elapsed)
+  input$target_daily_std <- (target$daily - mean(target$daily)) /
+    sd(target$daily)
+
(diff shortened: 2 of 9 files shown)


R/get_uptake_data.r
```diff
@@ -0,0 +1,153 @@
...
+    }
+  } else {
+    date_col <- rep_len(date, nrow(data))
+  }
+  date_col <- as.Date(date_col, format = date_format)
+
+  elapsed_col <- as.numeric(date_col - as.Date(start_date,
+                                               format = date_format))
+
+  year_col <- as.numeric(format(date_col, "%Y")) -
```
swo, line 131: is the idea here to get a "season"? Like, 2023/2024 season runs July 2023 to Aug 2024?

data/State_Name_Key.csv
swo, line 1: This is in base R under `state.name` and `state.abb`

Also, if you want to be a real R nerd, data like this should be either in `inst/` or in .rds files. See https://r-pkgs.org/data.html

scripts/projection_script.r
```diff
@@ -0,0 +1,65 @@
+setwd("/home/tec0/covid_booster_uptake")
```
swo, line 1: Remove this line: it's user specific and unnecessary... this should run from any place on your machine!

scripts/projection_script.r
```diff
@@ -0,0 +1,65 @@
...
+source("get_uptake_data.r")
```
swo, line 2: Rather than all this, you should just `library(name_of_your_package)`!

scripts/projection_script.r
```diff
@@ -0,0 +1,65 @@
...
+
+# Load 2022 NIS data for USA
+nis_usa_2022 <- get_uptake_data(
+  "https://data.cdc.gov/api/views/akkj-j5ru/rows.csv?accessType=DOWNLOAD",
+  state = "Geography",
+  date = c("Time.Period", "Year"),
+  cumulative = "Estimate....",
+  state_key = "USA_Name_Key.csv",
+  date_format = "%m/%d/%Y",
+  start_date = "9/2/2022",
```
swo, line 29: if *we* are manipulating dates, we always use https://en.wikipedia.org/wiki/ISO_8601, ie yyyy-mm-dd

the workflow should always be to take whatever the input date format is, and get it to ISO 8601 (or to date objects) as soon as possible

swo: oh, I see, you might be doing that below

swo (approved):
I'm marking as "approve," because I think the main goal of this PR is to get the code in. We can tweak stuff later, but even that isn't critical, if we're going to port it.

I have a lot of small comments about R coding practice. I think the biggest points are:

- Prefer `map` to loops
- Don't be afriad to use tidyverse (eg `dplyr::lag`, tibble manipulations) when it will make things clearer
- Use `lubridate` for date manipulation

If you haven't made a true R package before, I think this is also a good time to do that! Maybe in another PR. `usethis::create_package()` makes that very easy and gets your set up. You're missing a `DESCRIPTION`, `NAMESPACE`, and test framework.

It's also worth thinking about very clearly separating model construction, fitting, prediction, and evaluation. (R tends to mush construction and fitting together, see https://scikit-learn.org/stable/auto_examples/linear_model/plot_ols.html#sphx-glr-auto-examples-linear-model-plot-ols-py for how python keeps these things separate.) That will make it easier to generalize your code to different evaluation metrics.

Happy to talk about any of this sync; I know there's a lot here.

eschrom:
Thanks for all these comments, @swo. My answer to most of them is "Oh, I didn't know you could ..." I only know a few tricks in dplyr, I have never used lubridate, never worked with S3 classes, etc., hence my rudimentary approaches. I will only comment back on the rare specifics where there is a reason for my approach.

But overall I agree it's best to just move past this v1 and into Python. For smart architectures (e.g. separating model construction from fitting) and proper package organization, I should just learn in Python as soon as possible, rather than fussing with a soon-to-be-defunct R version.

R/build_projection_model.r
```diff
@@ -0,0 +1,64 @@
...
+#'
+#' @return a brms model object
+#'
+#' @export
+#'
+build_projection_model <- function(data, drop_first = NULL) {
+
+  if (is.null(drop_first)) {
+    intervals <- diff(data$elapsed)
+    if (abs(data$elapsed[1] - mean(intervals)) > (sd(intervals) + 1)) {
```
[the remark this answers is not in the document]
eschrom: In `get_uptake_data.R`, the user defines a `start_date` which is the rollout date. `elapsed` is the number of days since that rollout date. Generally the first data point is after the rollout date, and sometimes by a lot. Thus, the first `elapsed` represents a time interval perhaps longer than most other intervals (which are often 7 days, but sometimes 6 or 8). I want to make sure the first interval is not much more than 1 standard deviation bigger than the others. This condition naturally becomes less strict if there is a lot of variation in the intervals (which could happen in other data sets, though not the data sets we are working with currently). But the opposite concern is if all the intervals are exactly 7 days, for example. Then, sd = 0, so if the first interval is 6 or 8 days, you still don't use it, which may not be reasonable. The "sd + 1" builds in a little tolerance for this.

scripts/projection_script.r
eschrom: Yep, my mistake. I actually knew this one and just forgot!

R/evaluate_projection.r
```diff
@@ -0,0 +1,215 @@
...
+#' the projections will be compared against that. If new data, i.e. data
+#' which the model has not seen, is also provided, projections will compared
+#' against that instead. In the latter case, the fit may not be as good, and
+#' credible intervals will be wider due to extra uncertainty over the values
+#' of year-specific parameters.
+#'
+#' If no start date is provided, it is assumed the exact incident and
+#' cumulative uptake is known for each time t-1 when incident uptake for
+#' time t is being predicted. This evaluates how well the model fits
+#' each individual data point through the entire time series.
```
Fuhan-Yang, line 20: I'm not sure if I understand this. How can we predict without a start date?

R/evaluate_projection.r
```diff
@@ -0,0 +1,215 @@
...
+  output$incident_lower <- rep(NA, nrow(output))
+  output$cumulative_mean <- rep(NA, nrow(output))
+  output$cumulative_upper <- rep(NA, nrow(output))
+  output$cumulative_lower <- rep(NA, nrow(output))
+
+  if (!is.null(start_date)) {
+    start_date <- as.Date(start_date)
+    early_dates <- which(output$date <= start_date | is.na(output$previous))
+  } else {
+    early_dates <- which(is.na(output$previous))
```
Fuhan-Yang, line 100: So `early_dates` is not actual date but index. And the entire `output` will be used for evaluation, i.e, the `start_date` will be the start of the `output`, if there is no `start_date`. It may help to read if adding some comments there.

R/get_uptake_data.r
```diff
@@ -0,0 +1,153 @@
+#' Format uptake data
+#'
+#' Download a local or web .csv file of cumulative uptake percentages.
+#' You must name the columns that contain the state, date, and uptake.
+#'
```
Fuhan-Yang, line 5: Just thinking, is there an API for the data we need? Then we can automatically fetch the data instead of manually downloading! I'm new to getting data from API and would like to know more!

scripts/projection_script.r
```diff
@@ -0,0 +1,65 @@
+setwd("/home/tec0/covid_booster_uptake")
+source("get_uptake_data.r")
+source("interpolate_dates.r")
+source("build_scaling_model.r")
+source("evaluate_scaling.r")
+source("build_projection_model.r")
+source("evaluate_projection.r")
+source("projection_functions.r")
+
```
Fuhan-Yang, line 9: You can also use `devtools::load_all(the_directory_with_all_source_files)` or `load_code()`

scripts/projection_script.r
```diff
@@ -0,0 +1,65 @@
...
+# Load 2022 IIS data for USA
+iis_usa_2022 <- get_uptake_data(
+  "https://data.cdc.gov/api/views/unsk-b7fc/rows.csv?accessType=DOWNLOAD",
```
Fuhan-Yang, line 12: This is reading from API! Ignore my previous comments.

Fuhan-Yang (approved):
This is great! The code is clear and readable to me. I left some comments for small changes. I see the retrospective evaluation is also included here. We can talk about our next steps after merging.

R/build_projection_model.r
```diff
@@ -0,0 +1,64 @@
...
+    data <- data[-1, ]
+  }
+
+  data$previous <- c(NA, data$daily[-nrow(data)])
+  data <- data[-1, ]
+  data$previous_std <- (data$previous - mean(data$previous)) / sd(data$previous)
+  data$elapsed_std <- (data$elapsed - mean(data$elapsed)) / sd(data$elapsed)
+  data$daily_std <- (data$daily - mean(data$daily)) / sd(data$daily)
+
+  model <- brms::brm(daily_std ~ previous_std * elapsed_std,
```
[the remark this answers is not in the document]
eschrom: That, but with an intercept: X_t = a + b * X_t-1 + c * t + d * X_t-1 * t

R/build_projection_model.r
```diff
@@ -0,0 +1,64 @@
...
+#'
+#' BEWARE: THIS DOES NOT YET ACCOUNT FOR MULTI-YEAR OR MULTI-STATE DATA SETS
+#'
+#' Future incident uptake can be predicted from current incident uptake.
+#' A model can be built and fit to project incident uptake into the future.
+#'
+#' The default model says that current incident uptake is a linear function
+#' of the previous incident uptake, time elapsed since rollout, and their
+#' interaction.
+#'
```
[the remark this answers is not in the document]
eschrom: Elapsed is the time (in days) since the previous report. Incident is the change in uptake (in population %) since the previous report. Daily is the change in uptake (in population %) _per day_ since the previous report.

So daily = incident / elapsed.

R/build_scaling_model.r
```diff
@@ -0,0 +1,89 @@
...
+    model <- brms::brm(target_daily_std ~ daily_std * elapsed_std,
+                       data = input,
+                       family = gaussian())
+  } else {
+    model <- brms::brm(target_daily_std ~ daily_std * elapsed_std +
+                         (1 + daily_std | state),
+                       data = input,
+                       family = gaussian())
+  }
+
```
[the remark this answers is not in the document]
eschrom: Correct! `1 + daily_std | state` means that both the model intercept and the model slope for `daily_std` can vary by state.

R/evaluate_projection.r
eschrom: If no start date is given, each data point is predicted as if the previous data point were the start date. In other words, each data point is predicted as if the previous data point is known. This is useful for diagnosing the model on its training data, not for making bona fide projections of the future.

R/evaluate_projection.r
eschrom: Yes, I apologize that this is sloppy in the same way that many of @swo's comments point out sloppy or inefficient things!

eschrom:
Thank you, @Fuhan-Yang! While `evaluate_projection.R` does plot projected vs. realized uptake, there is no real retrospective evaluation in terms of quantitative scoring, like what you've been working on.

PR #2, opened 2024-09-18
Created a mostly-empty incident_regression python package

eschrom:
I used poetry to create the scaffold for a (currently empty) python package, with just a few dependencies that I know of so far - polars, datetime, and numpy. This is where we want to focus our effort on building well-structured code for uptake projections based on incident uptake regression models.


eschrom:
Regarding `poetry new` from within the existing repo, that is exactly what happened. I will try again. Let's go with `iuf` since it matches the name of the repo. New attempt incoming soon!

swo (approved):
lgtm!

PR #4, opened 2024-09-23
First Python version with simple architecture

eschrom:
This is the Python version of my original R, consisting of three functions (in /iup) and a script that calls them sequentially (in /scripts). Either IIS or NIS data is imported and formatted, a linear regression model is fit to the per-day incident uptake, and projections from this linear model are generated. Projections are returned as a DataFrame in the same format as the original data.

This code is meant to be minimal. Many features (some of which were present in the R version) are omitted here, e.g.:
- scaling between data sets
- interpolation of dates
- ability to specify any model other than:   uptake ~ 1 + time_elapsed + previous_uptake + time_elapsed*previous_uptake
- group-level variation on state or calendar year
- quantification of projection uncertainty
- any graphical anything

The purpose of this PR is to get a minimal Python version in our shared space. That way, @Fuhan-Yang has a starting point for building evaluation metrics in Python, while I can build in more features and smarter architectures with the help of @swo.

### DIFF
--- a/iup/build_projection_model.py
+++ b/iup/build_projection_model.py
@@ -1,0 +1,54 @@
+import polars as pl
+from sklearn.linear_model import LinearRegression
+
+
+def build_projection_model(data):
+    """
+    Fits a linear regression model to explain per-day incident uptake.
+
+    Parameters
+    ----------
+    data : DataFrame
+        Uptake data as imported and formatted by get_uptake_data.py
+
+    Details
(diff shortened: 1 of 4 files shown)


Fuhan-Yang (commented):
The description is clear in showing the use of the functions and the script. I don't have much to say about the coding since I'm still learning. Thank you Ed for putting this together!

PR #5, opened 2024-09-26
Retrospective evaluation - basic R version

Fuhan-Yang:
Hello @swo and @eschrom !

This is a basic R version of plotting and evaluating the forecasts initiated by multiple dates. Please check the commit `retrospective evaluate basic R version`. I also changed the original `build_projection_model.R` by adding the option to plot, and moving the data standardization into a small function. Please find the details in the code. Feel free to comment as much as you can! Inputs on function structure, coding habits are appreciated!


R/generate_projections.R
```diff
@@ -0,0 +1,207 @@
...
+    pred_summary <- data.frame(
+        date = test_data$date,
+        obs = test_data$daily,
+        lower = apply(pred, 2, quantile, 0.025),
+        upper = apply(pred, 2, quantile, 0.975),
+        mean = colMeans(pred)
+    )
+
+    # predicted cumulative vaccine uptake #
+    cumu <- t(apply(pred, 1, cumsum))
```
eschrom, line 94: Cumulative sums of incident uptake alone will be too small, because we also need to include the cumulative uptake from the first 1 or 2 data points of the season, which were dropped prior to model fitting.

R/generate_projections.R
```diff
@@ -0,0 +1,207 @@
...
+    ## recover the prediction to the original format for comparision ##
+    pred <- inv_z_scale(test_data$daily, scaled_pred)
+
+    # predicted daily vaccine uptake #
+    pred_summary <- data.frame(
+        date = test_data$date,
+        obs = test_data$daily,
+        lower = apply(pred, 2, quantile, 0.025),
+        upper = apply(pred, 2, quantile, 0.975),
+        mean = colMeans(pred)
```
eschrom, line 90: This is the per-day average incident uptake over the time period between data points (typically 7 days). It must be multiplied by the time interval since the previous data point (in days) to be the actual incident uptake.

eschrom (commented):
Thanks for doing all this work, Fuhan!! I suspect there are flaws regarding data processing (dropping dates and standardizing variables), bookkeeping with per-day incident vs. incident vs. cumulative uptake, and the process of forward projection with appropriate uncertainty. Please see my individual comments for details.

If I am wrong in any of my comments, please reply to them to correct me! If I am right, perhaps you still don't want to invest the time in fixing them in this R code, since we already have a Python version started. You could instead just keep this PR to reference as you add to the initial Python version. 

Let me know if you want to discuss anything further!

scripts/evaluation_script.r
```diff
@@ -0,0 +1,56 @@
...
+    cumulative = "estimate",
+    state_key = "data/USA_Name_Key.csv",
+    date_format = "%m/%d/%Y",
+    start_date = "9/12/2023",
+    filters = list(c("time_type", "Weekly"), c("group_name", "Overall"))
+)
+
+
+# generate projections initiated at different time #
+all_data <- rbind(nis_usa_2022, nis_usa_2023)
```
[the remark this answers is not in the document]
Fuhan-Yang: I agree that this data input is not practical, I did this is to have more data to train. This may cause the issue of 'drop_first' too. Maybe it's worth to put a warning/error sign when the data of multiple seasons is used in python, as in most cases we will only project within one season.

R/generate_projections.R
```diff
@@ -0,0 +1,207 @@
...
+#' @param test_data: data frame used to compare with the prediction generated by linear regression
+#' @return a list containing the predicted uptake and the corresponding
+#' observed rate under the `test_data` time frame,
+#' in both uptake rate `rate` and cumulative uptake `cumulative`
+
+real_time_projection <- function(
+    train_data = nis_usa_2022,
+    test_data = nis_usa_2023) {
+    # transform the train data and test data to the format of model building and model prediction #
+    scaled_train <- data_prepare(train_data)
```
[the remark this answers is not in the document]
eschrom: I've built this logic into the python version, so hopefully you won't have to worry about it!

R/generate_projections.R
```diff
@@ -0,0 +1,207 @@
...
+    scaled_test <- data_prepare(test_data)
+
+    # The first row is also removed for raw test data, to be comparable with the predicted values #
+    test_data <- test_data[-1, ]
+
+    # model building #
+    model <- build_projection_model2(data = scaled_train)
+
+    ## get out-of-sample prediction ##
+    scaled_pred <- brms::posterior_predict(model, newdata = scaled_test)
```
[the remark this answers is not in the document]
eschrom: I also use brms, but I use `posterior_predict` to generate a posterior distribution for just one timepoint at a time. That posterior distribution is then un-z-scaled and re-z-scaled to become the `newdata` for the next timepoint.

R/generate_projections.R
```diff
@@ -0,0 +1,207 @@
...
+
+    ## recover the prediction to the original format for comparision ##
+    pred <- inv_z_scale(test_data$daily, scaled_pred)
```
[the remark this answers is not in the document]
eschrom: That's not quite what I mean. The means and standard deviations used for z-scaling (i.e. "standardization") should remain constant. In the `test_data` used to fit the model:

- the `previous` column has a mean "prv_mn" and standard deviation "prv_sd." The column is standardized such that `previous_std <- (previous - prv_mn) / prv_sd`
- the `daily` column has a mean "dly_mn" and standard deviation "dly_sd." The column in standardized such that `daily_std <- (daily - dly_mn) / dly_sd`

`brms::posterior_predict` gives a posterior predictive distribution for the first future unknown timepoint. Let's call it "ppd_1." 
- To put "ppd_1" on the natural scale, we must `ppd_1_nat <- ppd_1 * dly_sd + dly_mn`
- To use "ppd_1" as "previous" input to get a posterior distribution for the second future unknown timepoint, we must further `ppd_1_prv <- (ppd_1_nat - prv_mn) / prv_sd`

Now `ppd_1_prv` can be used in `brms::posterior_predict` to get `ppd_2` and so on. The process continues to iterate, but the constants used for standardization ("prv_mn," "prv_std," "dly_mn," & "dly_sd") never change - they are fixed by the data used to fit the model.

scripts/evaluation_script.r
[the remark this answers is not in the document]
eschrom: I think we will often train the model based on multiple seasons of past data, but we will only project within one season. So maybe your sliding start date for the projection evaluations should only slide within the most recent season represented in `all_data` and never reach back to past seasons. Do you think that is a reasonable approach?

R/generate_projections.R
```diff
@@ -0,0 +1,207 @@
...
+    all_data, min_data_size = 8) {
+    start_date <- sort(all_data$date)[1 + min_data_size - 1]
+    end_date <- sort(all_data$date)[nrow(all_data) - min_data_size + 1]
+
+    date_series <- all_data %>%
+        filter(date >= start_date, date <= end_date) %>%
+        select(date)
+
+    # convert single-column data.frame to vector #
+    date_series <- as.vector(date_series$date) + as.Date("1970-01-01")
```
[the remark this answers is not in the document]
eschrom: Ha! I wonder why Jan 1, 1970 was chosen. In any case, thanks for filling me in!

R/generate_projections.R
```diff
@@ -0,0 +1,207 @@
...
+        upper = apply(pred, 2, quantile, 0.975),
+        mean = colMeans(pred)
+    )
+
+    # predicted cumulative vaccine uptake #
+    cumu <- t(apply(pred, 1, cumsum))
+    cumu_summary <- data.frame(
+        mean = colMeans(cumu),
+        obs = test_data$cumulative,
+        lower = apply(cumu, 2, quantile, 0.025),
```
swo, line 98: This matrix stuff makes me so crazy:

- `posterior_predict()` returns an array with S (number of posterior draws) rows and N (number of data points) columns; each cell is an incident uptake
- `apply(pred, 1, cumsum)` takes the cumulative sums over each row, giving a new S x N matrix; each cell is a cumulative uptake; you then need to `t()` that?
- `apply(cumu, 2, quantile, ...)` takes the quantiles over each column (rows are samples so this gives you a quantile over samples)

The better way to do this is to have every model return a data frame with a consistent schema. I think you want columns: date, posterior draw ID, and incident uptake.

Then, you can have post-processing like this function that is common to all models, and which can use tidyverse, not matrix, operations:
```r
cumu <- model_output |>
  arrange(draw_id, date) |>
  group_by(draw_id) |>
  mutate(cum_uptake = cumsum(inc_uptake)) |>
  ungroup()

cumu |>
  group_by(date) |>
  summarize(
    cum_uptake_median = median(cum_uptake),
    cum_uptake_lci = quantile(cum_uptake, 0.025)
  ) |>
  left_join(test_data, by = "date")
```

R/generate_projections.R
```diff
@@ -0,0 +1,207 @@
...
+
+#' @description This function plots prediction with observed data
+#' This function is embeded in `plot_real_time_projection`
+#' @param option: which uptake to plot: 'rate' or 'cumulative'
+#' @param predicted: out-of-sample prediction from linear regression
+#' @param test: the observed data corresponding to the prediction
+#' @return a ggplot object
+
+plot_projection <- function(option, predicted, test) {
+    if (option == "rate") {
```
swo, line 114: The more elegant way to do this is with https://dplyr.tidyverse.org/reference/case_match.html:

```r
data_y_name <- case_match(
  option,
  "rate" ~ "daily",
  "cumulative" ~ "cumulative"
)
if (is.null(data_y_name)) {
  stop("whatever")
}
```

R/generate_projections.R
```diff
@@ -0,0 +1,207 @@
...
+#' @param all_data: Data used to iteratively train and test,
+#' with each iteration moving forward one data point
+#' @param min_data_size: the minimum data size required for model training,
+#' So far, it is the same for model testing.
+#' @return a X*Y grid ggplot
+
+plot_real_time_projection <- function(
+    data_option = "rate",
+    all_data, min_data_size = 8) {
+    start_date <- sort(all_data$date)[1 + min_data_size - 1]
```
swo, line 163:
```suggestion
    start_date <- sort(all_data$date)[min_data_size]
```

R/generate_projections.R
```diff
@@ -0,0 +1,207 @@
...
+    end_date <- sort(all_data$date)[nrow(all_data) - min_data_size + 1]
+
+    date_series <- all_data %>%
+        filter(date >= start_date, date <= end_date) %>%
+        select(date)
```
swo, line 168: I think you want `pull()` here? to get a vector

R/generate_projections.R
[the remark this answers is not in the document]
swo: Strongly recommend against this kind of conversion.

See note above, the easiest thing is to do is to just `pull()`, to get a vector of date objects

Whenever working with dates or times in R, always always look to https://lubridate.tidyverse.org/

It's way too easy to make mistakes with dates

R/generate_projections.R
```diff
@@ -0,0 +1,207 @@
...
+
+    # convert single-column data.frame to vector #
+    date_series <- as.vector(date_series$date) + as.Date("1970-01-01")
+    ps <- list()
+    i <- 1
+
+    for (split_date in date_series) {
```
swo, line 175: Any time you have a for loop, instead make it a function!

```r
plot_forecast <- function(split_date) {
  # everything you have in this loop, ending with the plot
}

ps <- map(date_series, plot_forecast)
```

R/retrospective_evaluate.r
swo, line 1: I see a bunch of "don't repeat yourself" problems between this file and the one before. I don't need to be super-picky about this, as we're about to port to Python, but your code should allow you to fit each model exactly once.

Because the data are so small, it would make more sense to have a function that takes the model function and the data, then splits into training and eval for each forecast date, fits a model for each forecast date. Then you can have some other function that knows how to take a list(training_data, eval_data, model_output) and turn that into a plot, or diagnostic data, or predictions, or whatever you want.

data/NIS_2023-24.csv
swo, line 1: This is data and should not be in the repo

data/NIS_2023-24.csv:Zone.Identifier
swo, line 1: This is a nuisance file and should not be in the repo

This is normally a sign that you've copy-pasted files between Windows and Linux

You can add this to your user-global .gitignore

iup/__pycache__/build_projection_model.cpython-311.pyc
swo, line 1: This repo should have a .gitignore that ignores .pyc files

iup/get_real_time_projections.py
swo, line 1: I'm confused by this PR: it says "basic R version," but it also has python in it? I think these got submitted to this branch when you didn't mean them to; they should be taken out of this PR and reviewed separately

scripts/evaluation_script.r
```diff
@@ -0,0 +1,56 @@
+rm(list = ls())
+# read all the R files under R folder #
+R_path <- "R/"
+purrr::map(paste0(R_path, list.files("R/")), source)
```
swo, line 4: We should never `source()` files in our scripts

The solutions are to have the user `devtools::load()` or to just install the package

Depending on how quickly we are going to trash the R code, I will be more or less picky about this. But in general this is a coding practice that will quickly get you into a lot of trouble, when you start to have conflicting names, etc.

scripts/evaluation_script.r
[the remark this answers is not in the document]
swo: I have the sense that Ed's model can only be trained on one season at a time, ie there is no pooling across seasons, because there is no hierarchy? (Or, if it *does* have hierarchy, we shouldn't be starting from there; we should be starting from models that treat each season independently.)

scripts/projection_script.py
swo, line 1: Also belongs in a separate PR

data/NIS_2023-24.csv
swo: eh, OK, actually, it's OK to have this in the repo, since it's so small and we don't have other tools (eg S drive or Azure) set up to do this exchange

but this should be in a separate PR, not here

tests/test.py
swo, line 1: This belongs in a separate PR

Also, this appears to be a work in progress; it's not a working test suite

If this is a scratch file, it should go into a `tmp/` folder or something similar. The folder `tests/` has a very specific structure and function

swo (changes requested):
In order of how gating my comments are, to landing this PR:

1. There are files in this branch that shouldn't be committed to Git at all; those should be removed and you should set up gitignore's to avoid in the future
2. There are Python files in this PR that belong in a separate PR for that porting
3. There is a data file that can be in this repo, but that should also be in a separate PR, so we can track when we do data updates
4. Ed had some substantive things about how the code actually works. I leave it up to the two of you whether you want to fix all those things in this PR, or just say "we know this stuff isn't exactly right, but we're going to work on the Python version rather than fix this"
5. I have a lot of style/efficiency things that can make your R code cleaner and more up to CFA standards. Those are mostly for your education at this point

R/generate_projections.R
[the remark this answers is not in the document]
Fuhan-Yang: A similar version of these functions is `inv_z_scale()` defined at line 31-35. These functions are created and their names are clarified in the python version.

R/generate_projections.R
Fuhan-Yang: Daily prediction is used in the python version.

R/generate_projections.R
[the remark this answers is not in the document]
swo: My point is that your `inv_z_scale()` takes two arguments: the vector that's used to derive the reference mean & SD, and then the vector to be inverse-scaled

Ed's point is that you may not have the same reference data after you do some of the data manipulation in the middle of that function

So my point is that define functions of *one* argument, the vector to be scale or inverse-scaled, and use those before and after the data fitting, so you can be sure you're using the same reference values in both places

R/generate_projections.R
swo: But why *don't* we do weekly? We'll never have daily data, so why are we trying to interpolate down to days?

eschrom: Because the intervals are not always a clean 7 days. They are usually 7 days, but there's an occasional 6 or 8 thrown in. Interpolating down to days allows us to overlook this without worrying that 6-day intervals will have slightly smaller incident uptake (or 8-days slightly larger) than the customary 7-day interval would.

scripts/evaluation_script.r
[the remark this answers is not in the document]
eschrom: No hierarchy at the moment - I would just recommend we keep that possibility in mind for the future

R/generate_projections.R
swo: Understood. How do you interpolate down to days? Is there some weird statistical artifact that would come from assuming constant uptake over a week?

The solution in the Bayesian world is nicer I think. We posit that there is an underlying (but unobserved) daily incident uptake, and the thing we observed is the cumulative uptake as of certain time points (or the incident uptake over some periods of time, which can be of arbitrary duration and different from one another)

eschrom: Linear interpolation. Given that the entire uptake curve is nonlinear, linear interpolation is surely "wrong," as you suspect. But on the few example data sets I've used, the posterior predictive checks look great, so I don't think it is "wrong" enough to matter.

Your suggested approach assumes a daily incident uptake: is it a problem that the uptake isn't linear over the course of the day? What if uptake peaks just before lunch and then again mid-afternoon? I see it like calculus - we can't ever quite escape the problem of nonlinearity in time, but we can squeeze it into smaller and smaller intervals until the problem is negligible. If 6-8 days is a small enough interval to make the problem negligible, I'm happy with the current approach. If we need smaller 1-day intervals, I'm happy to switch to your approach - I would just need to learn a bit more about it!

swo:
@eschrom There are some things that still need to get done. My recommendation to @Fuhan-Yang was to open 3 new PRs, and eventually close this one:

1. R code only (basically this branch, but without the Python, data, or other files that should be gitignore'd)
2. Data (I'm not super-opposed to having small data files in this repo, if there's no other place to get them, but it should be a separate PR)
3. New Python stuff -- and probably only one or two functions there

> In order of how gating my comments are, to landing this PR:
> 
> 1. There are files in this branch that shouldn't be committed to Git at all; those should be removed and you should set up gitignore's to avoid in the future
> 2. There are Python files in this PR that belong in a separate PR for that porting
> 3. There is a data file that can be in this repo, but that should also be in a separate PR, so we can track when we do data updates
> 4. Ed had some substantive things about how the code actually works. I leave it up to the two of you whether you want to fix all those things in this PR, or just say "we know this stuff isn't exactly right, but we're going to work on the Python version rather than fix this"
> 5. I have a lot of style/efficiency things that can make your R code cleaner and more up to CFA standards. Those are mostly for your education at this point

R/generate_projections.R
swo: I'm more concerned about statistical anomalies. Like, by going from 1 data point to 7, are you changing the confidence intervals somehow? Are you going to produce some weird edge effect (that won't be undone by summing daily uptakes into weeks after the analysis)?

This kind of interpolation will also be tricky when we do the v2 treatment, because although I know the observation error on the weekly data (ie the reported CIs), but I don't the error on the interpolated daily uptake

eschrom: 1 data point is not becoming 7 data points; 1 data point stays 1 data point. That data point is simply normalized. If the incident uptake for the 7-day period ending on 2024-10-17 is 7%, then the daily average incident uptake for the 7-day period ending on 2024-10-17 is 1%. It's still a single data point on a single date - it's just normalized in a sense.

swo:
Close in favor of #15 , #13 , #12

PR #9, opened 2024-10-17
Evaluation for R

Fuhan-Yang:
Hi Scott @swo and Ed @eschrom, this branch is inherited from `fy_eval` with R files only. No need to review the code as it was the same as in `fy-eval`, and the suggested changes have been made in the python branch `fy_eval_py`.


swo:
Close in favor of #15

PR #10, opened 2024-10-17
Create python branch

Fuhan-Yang:
Hi Scott@swo and Ed @eschrom , this branch is inherited from `fy_eval` but with updated python files. The python file is the first version of generating vaccine uptake projections. Please leave any comments, then I can integrate in the plotting and evaluation modules in the next steps.


swo:
There are still a bunch of files that shouldn't be in this PR, like the Zone Identifier, etc.

Also, this PR both adds the Python stuff *and* removes the R stuff, which may or may not have been your intent

swo:
Close in favor of #13

PR #11, opened 2024-10-17
Create a branch for data upload only

Fuhan-Yang:
Hi Scott @swo and Ed @eschrom, this is the branch for uploading the data only. It is rare to upload data in github, while the data size is small, so this PR is just used as an example of uploading data in the workflow, if needed in the future.


swo:
I think you want to rebase this branch and keep only https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/11/commits/bdaa9ad9be5a17238672f8dda0e2333740921e3d . Because if you look at Files now, you'll see that this PR deletes a bunch of files, when all you want to do is to add the data

swo:
See #12 for what I think this PR should look like, although even that one is missing documentation. Feel free to open a branch off of #12, add some documentation, and we can merge in there!

swo:
Closing in favor of #12

PR #14, opened 2024-10-17
Add Python gitignore

### LINKED ISSUE #8: Add gitignore
If we're going to end up using python, use this one: https://github.com/github/gitignore/blob/main/Python.gitignore

swo:
Standard Python gitignore, with .vscode/ added

Resolves #8

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,0 +1,202 @@
+# Data
+*.csv
+*.tsv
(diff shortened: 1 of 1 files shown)


Fuhan-Yang (approved):
Thanks for resolving this, Scott!

PR #19, opened 2024-10-17
Remove issue templates

swo:
@Fuhan-Yang @eschrom If either of you like the issue templates, we can keep them, but I find that the existing templates only add friction


Fuhan-Yang:
This works for me!

Fuhan-Yang (approved):
I'm ok with removing the issue templates!

PR #23, opened 2024-10-22
add plotting

Fuhan-Yang:
Hi Scott and Ed, 

This branch is developing to plot the weekly-initiated predictions with the observed data. Comments are welcome!


iup/plot_projections.py
```diff
@@ -0,0 +1,96 @@
...
+        The name of the graph.
+    
+    Return:
+    -------------
+    None. The graph is saved. 
+
+    """
+    
+    # input check #
+    if 'date' not in obs_data.columns or 'cumulative' not in obs_data.columns:
```
swo, line 35: See #7 for some examples of how to move data validation into data objects

iup/get_projections.py
```diff
@@ -198,12 +198,17 @@ def get_projection(
...
                 daily = pl.lit(fitted_daily)
+        ).with_columns(
+            cumulative = pl.col('cumulative') + pl.col('previous')
         )
 
         pred_data = pl.concat([pred_data, 
                             raw_data],
                                 how = 'vertical')
 
+    pred_data = pred_data.select(
```
swo, line 209: Try chaining this with the above! the `x = x.something()` paradigm is much harder to reason about than a chain like `pred_data = pl.concat([whatever]).select([whatever])`

iup/get_projections.py
swo, line 1: The R paradigm is to have ~1 function per code file. E.g., you have `get_projections.R` and that file has the function `get_projections()` and not much else.

If you look at Python codebases, the preference is to group code much more. For this project, you can probably keep all the Python code in the `__init__.py`. Or, at least, combine the code in this file and in `plot_projections.py`, both of which do evaluation or viz stuff (and so could be in an `evaluation.py` or similar)

swo:
See #27 for some other changes that will be important before we can merge

swo (approved):

PR #24, opened 2024-10-23
Add data description

Fuhan-Yang:
Hi Scott and Ed, 

This is an atomic PR to add the data description in the README.md and delete the `NIS_2023-24.csv`, which can be acquired through API. 

This is the public API that returns the same data as in the csv: "https://data.cdc.gov/resource/pakc-hru3.json". The `get_uptake_data()` currently does not have the `.json` input option, maybe this can be added in the future development.


swo (changes requested):
Before anything else, I think you need to rebase? If you look, there are 12 files changed in this PR, including the github issue templates!

Fuhan-Yang:
It should work now!

swo (approved):

Issue #33, opened 2024-10-25
Separate `get_nis` from `parse_nis`


eschrom:
Should the acquisition of a data set from a file system or link be its own function, followed by a separate function to pre-process and format the data set? In this case, `parse_nis` would be the place to write a suite of functions tailored to different NIS data sets. However, if the only advantage of separating `get_nis` from `parse_nis` is so that acquiring the data set (via internet connection) isn't included in the test, this could alternatively be solved by only writing tests that use small mock data sets stored in the repo.

Issue #29, opened 2024-10-25
Reformat season in uptake data objects


eschrom:
Season should include both the fall and spring years (for overwinter seasons, that is). E.g. "2024/25" not "2024"
Also, consider making season something that the data must already contain upon retrieving, rather than manipulated in the code.

swo:
Hopefully "season" will be unambiguous for the things we're dealing with. CDC defines "season" differently for different pathogens and surveillance systems (including some that have an interseason period).

For something like season flu and COVID vaccinations we should be able to do something simple that is unambiguous

PR #34, opened 2024-10-25
Next Attempt at Object-Oriented Architecture

eschrom:
Comments on my first attempt at an object-oriented architecture (#6) warranted a total overhaul. Here is my new attempt. Many of the comments from #6  are addressed, and the rest I recorded as issues (not linked to PR#6, though, because I closed that). Plenty of improvement is still necessary, but I think I'm ready for the next round of feedback.


iup/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+        ).all(), "Cumulative uptake estimates cannot be negative."
+
+    def to_incident(self) -> IncidentUptakeData:
+        self = (
+            self.with_columns(
+                season=(
+                    pl.col("date").dt.year()
+                    + pl.when(pl.col("date").dt.month() < 7).then(-1).otherwise(0)
+                ).cast(pl.Utf8),
+                estimate=pl.col("estimate").diff().over("region"),
```
swo, line 113: the "over regions" thing is giving me heartburn. Again, if we wanted to add/remove a column (eg population subgroup or demography), then we'd need to go look for the magic word "region" everywhere in the code

I'd much rather having something that's like "grouping variables" or something, and then you are always going "over" those variables, so you only need to have them written in one place

(I'll take this as another opportunity to say that I think it's easier to think about "uptake data" as being *only* dates and estimates. Then all these data frame manipulations become simpler, because you're only looking at a single time series for each one. For example, if you have columns region, season, date, estimate, then:

```
df.partition_by(['region', 'season'], as_dict=True, include_key=False)
```

would give you a dictionary `{(region, season): df}` where each little df has only date & estimate.

I should probably just open a PR to demo this, once this one gets landed.)

iup/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+                .fill_null(pl.col("estimate").first())
+                .over("region"),
+                interval=pl.col("elapsed").diff().over("region"),
+            )
+            .with_columns(
+                interval=pl.col("interval")
+                .fill_null(pl.col("elapsed").first())
+                .over("region")
+            )
+            .with_columns(estimate=(pl.col("estimate") / pl.col("interval")))
```
swo, line 130: does this mean that the incident "estimate" is daily uptake? I think this is why I prefer the idea of having start date, end date, and uptake during that period, because then it's clear that the NIS data is weekly uptake, while your daily model would want to extrapolate out to daily uptake

iup/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+            .with_columns(previous=pl.col("estimate").shift(1).over("region"))
+        )
+
+        self = IncidentUptakeData(self)
+
+        return self
+
+    @staticmethod
+    def split_train_test(uptake_data_list, start_date, side):
```
swo, line 139: This seems identical to the other function, above? That tells me it should be part of the parent class?

iup/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+    rollout,
+    filters=None,
+) -> CumulativeUptakeData:
+    out = (
+        (pl.read_csv(data_path))
+        .with_columns(estimate=pl.col(estimate_col).cast(pl.Float64, strict=False))
+        .drop_nulls(subset=["estimate"])
+    )
+
+    if filters is not None:
```
swo, line 172: This should be its own separate function, that deserves some unit testing!

iup/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+        filter_expr = pl.lit(True)
+        for k, v in filters.items():
+            filter_expr &= pl.col(k).is_in(pl.lit(v))
+        out = out.filter(filter_expr)
+
+    out = (
+        out.with_columns(
```
swo, line 179: ditto, this should also be broken out

iup/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+        )
+        .select(["region", "date", "estimate"])
+        .sort("region", "date")
+    )
+
+    unique_regions = out["region"].unique()
+    rollout_rows = pl.DataFrame(
+        {
+            "region": unique_regions,
+            "date": [rollout for _ in range(len(unique_regions))],
```
swo, line 193:
```suggestion
            "date": rollout,
```

iup/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+            "estimate": [0.0 for _ in range(len(unique_regions))],
```
swo, line 194:
```suggestion
            "estimate": 0.0
```

iup/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+
+        data = IncidentUptakeData(  # Not sure why I must reinforce type here
+            data.trim_outlier_intervals().with_columns(
+                previous_std=pl.col("previous").pipe(standardize),
+                elapsed_std=pl.col("elapsed").pipe(standardize),
+                estimate_std=pl.col("estimate").pipe(standardize),
+            )
+        )
+
+        self.previous_mean = data["previous"].mean()
```
swo, line 250: I think there's some clever way you could do this:
```python
self.standards = [
  getattr(data[x], y)()
  for x in ['previous', 'elapsed', 'estimate']
  for y in ['mean', 'std'])
]
```
although presumably you'd want that as some kind of dictionary

scripts/script.py
swo, line 1: let's give this a more informative name?

swo: also, it would be good for the README to say how to call this, what's the entry point into the repo

scripts/script.py
```diff
@@ -0,0 +1,43 @@
+import yaml
+import iup
+
+# Access the YAML config file
+with open("config.yaml", "r") as file:
+    config = yaml.safe_load(file)
+
+# List of the cumulative data sets described in the yaml
+cumulative_data = [
+    iup.get_nis(
```
swo, line 10: I'd suggest `iup.get_nis(**x) for x in config['data'].values()`!

scripts/script.py
```diff
@@ -0,0 +1,43 @@
...
+incident_test_data = iup.IncidentUptakeData.split_train_test(
+    incident_data, config["timeframe"]["start"], "test"
+)
+
+# Figure out how to initialize and fit models using the training data
+incident_model = iup.LinearIncidentUptakeModel().fit(incident_train_data).predict(
+    config["timeframe"]["start"],
+    config["timeframe"]["end"],
+    config["timeframe"]["interval"],
+)
```
swo, line 43: and some kind of printed output maybe??

swo (changes requested):
A very good start! My biggest concerns are (1) documentation, mostly overall and then for particular functions and (2) the tech debt around hard-coding "region" in a bunch of places.

Everything else are my implementation hypotheses that we can address in ongoing fixes, rather than in a first PR

iup/__init__.py
eschrom: Issue #33

iup/__init__.py
```diff
@@ -0,0 +1,358 @@
...
+        self.assert_columns_type(["estimate"], pl.Float64)
+        assert (
+            self["estimate"] >= 0.0
+        ).all(), "Cumulative uptake estimates cannot be negative."
+
+    def to_incident(self) -> IncidentUptakeData:
+        self = (
+            self.with_columns(
+                season=(
+                    pl.col("date").dt.year()
```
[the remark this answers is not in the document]
eschrom: Also season should include both the fall and spring years - Issue #29

iup/__init__.py
```diff
@@ -0,0 +1,361 @@
...
+            )
+
+        out = IncidentUptakeData(out)
+
+        return out
+
+    def get_previous_uptake(self) -> Self:
+        self = self.with_columns(previous=pl.col("estimate").shift(1).over("region"))
+
+        return self
```
eschrom, line 65: Pylance flags this as an error, saying:  Type "DataFrame" is not assignable to return type "Self@IncidentUptakeData"

I don't understand this error. `get_previous_uptake` is quite analogous to `trim_outlier_intervals` in that it performs some data frame manipulation on an IncidentUptakeData object and returns it as the same object type. Why is this a problem in `get_previous_uptake` but not in `trim_outlier_intervals`?

eschrom: As far as I can tell, a polars method (e.g. `with_columns`) will always return a polars dataframe, even if it was given a custom subclass of a polars dataframe. To prevent this, I had to define my own `with_columns` method that does what polars `with_columns` does but then returns whatever subclass it was given.

iup/__init__.py
```diff
@@ -147,50 +147,75 @@ def to_incident(self) -> IncidentUptakeData:
...
+    frame = fetch_nis(path)
 
     if filters is not None:
-        filter_expr = pl.lit(True)
-        for k, v in filters.items():
-            filter_expr &= pl.col(k).is_in(pl.lit(v))
-        out = out.filter(filter_expr)
+        frame = apply_filters(frame, filters)
+
+    frame = select_columns(frame, region_col, date_col, estimate_col, date_format)
```
swo, line 222: optionally you could chain with https://docs.pola.rs/api/python/stable/reference/dataframe/api/polars.DataFrame.pipe.html

scripts/script.py
```diff
@@ -36,8 +26,12 @@
...
 
 # Figure out how to initialize and fit models using the training data
-incident_model = iup.LinearIncidentUptakeModel().fit(incident_train_data).predict(
-    config["timeframe"]["start"],
-    config["timeframe"]["end"],
-    config["timeframe"]["interval"],
+incident_model = (
+    iup.LinearIncidentUptakeModel()
+    .fit(incident_train_data)
+    .predict(
```
swo, line 37: if the keywords here are `start`, `end`, `interval`, then you could do `.predict(**config['timeframe'])`

sometimes that's really slick to do `f(**my_dict)`, but other times it is better to be explicit with `f(arg1=my_dict['arg1'], arg2=my_dict['arg2'])`

config/config.yaml
```diff
@@ -0,0 +1,19 @@
+# The data sets to load and how to interpret them
+data:
+  data_set_1:
+    path: NIS_Flu_2023.csv
```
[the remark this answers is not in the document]
swo: Either cache or make a template

iup/__init__.py
```diff
@@ -0,0 +1,739 @@
...
+            names of columns for which to check type
+        dtype:
+            data type for each listed column
+        """
+        for col in columns:
+            assert (
+                self[col].dtype == dtype
+            ), f"Column {col} should be {dtype} but is {self[col].dtype}"
+
+    def with_columns(self, *args, **kwargs):
```
swo, line 58: This is an interesting point, that any polars verb will return a normal `pl.DataFrame`, rather than an instance of this object

Personally, I think that's OK, and the user could wrap `UptakeData(df)` any time they want to validate or assert that `df` is actually uptake data

Because otherwise we'd need to figure out some way to ensure that `.rename()`, `.filter()`, etc. all do the same thing

iup/__init__.py
```diff
@@ -0,0 +1,739 @@
...
+        Parameters
+        date_col: pl.Expr
+            column of dates
+
+        Returns
+        pl.Expr
+            column of the number of days elapsed since the first date
+
+        Details
+        Date column should be chronologically sorted in advance.
```
swo, line 127: this is something that you could put an assert statement in for

iup/__init__.py
```diff
@@ -0,0 +1,739 @@
...
+                .sort("date")
+                .filter(pl.col("date") >= start_date)
+            )
+
+        out = orig_class(out)
+
+        return out
+
+
+class IncidentUptakeData(UptakeData):
```
swo, line 173: I found in the fall virus work that it was good to have columns like `inc_uptake` or `cum_uptake`, to be absolutely crystal clear about what the "estimate" was

something for the backlog

swo (approved):

iup/__init__.py
eschrom: Issue #42

iup/__init__.py
eschrom: Issue #43

iup/__init__.py
eschrom: Issue #44

PR #37, opened 2024-10-29
Add evaluation metrics

Fuhan-Yang:
This is the last PR before building up object-oriented architecture. It calculates and plots evaluation metrics, including MSPE and end-of-season uptake.


iup/eval_projections.py
```diff
@@ -1,5 +1,86 @@
...
+    -------------
+    A polars.DataFrame with the MSPE and the date of initializing forecast. 
+    
+    """
+    
+    
+    mspe = data_df.join(
+            pred_df, 
+            on = 'date',
+            how = 'inner'
```
swo, line 35: this is a good place for validation. I assume you want 1:1 matching?

iup/eval_projections.py
```diff
@@ -1,5 +1,86 @@
...
+        ).with_columns(
+            spe = (pl.col(var) - pl.col(f"{var}_right"))**2
+        ).with_columns(
+            mspe = pl.col('spe').mean(),
```
swo, line 39: what happens if data are missing in either of the two data frames?

I think you might want to have some checks that the two data frames have the same dates, and only one of each date, eg

iup/eval_projections.py
swo, line 1: All these files will need to go through pre-commit compliance. You'll get most of the way there if you set up ruff auto-formatting in VSCode

iup/eval_projections.py
```diff
@@ -1,5 +1,86 @@
...
+        ).select(
+            'date','mspe'
+        )
+    
+    return mspe
+
+
+# get end-of-season total uptake #
+
+def get_eos(
```
swo, line 51: I would have expected that this function would have the same signature as the one before it. We'll want all our evaluation functions to have the same signature

scripts/evaluation_script.py
```diff
@@ -48,3 +53,71 @@
     pic_loc="output/",
     pic_name="weekly_predicted_uptake.png",
 )
+
+# evaluate time-varying MSPE #
+date_mspe = pl.DataFrame()
+
+for i in range(len(rt_pred_2023)):
```
swo, line 60: rework as a list comprehension:
```python
date_mspe = pl.concat([
  get_mspe(whatever)
  for whatever in so_and_so
])
```

iup/eval_projections.py
swo: also, this doesn't seem like a score function, per se? it just pulls out a particular data point

scripts/evaluation_script.py
```diff
@@ -48,3 +53,71 @@
...
+date_eos = pl.DataFrame()
+
+for i in range(len(rt_pred_2023)):
+    eos = get_eos(
+        rt_pred_2023[i],
+        var = 'cumulative'
+    ).with_columns(
+        date = rt_pred_2023[i].filter(
+            pl.col('date') == pl.col('date').min()
+        ).select('date')
```
swo, line 93: what's going on here? this is getting the start of the season date?

scripts/evaluation_script.py
```diff
@@ -48,3 +53,71 @@
...
+    x=alt.X('date:T',title='Date',axis=time_axis),
+    y=alt.Y('mspe',title='MSPE')
+).save(
+    'output/MSPE.png'
+)
+
+# evaluate predicted end-of-season uptake #
+date_eos = pl.DataFrame()
+
+for i in range(len(rt_pred_2023)):
```
swo, line 86: again, avoid loops where possible!

scripts/evaluation_script.py
```diff
@@ -48,3 +53,71 @@
...
+    )
+
+    date_eos = pl.concat([date_eos, eos],how = 'vertical')
+
+
+# plot predicted end-of-season uptake with data #
+obs_vac = nis_usa_2023.filter(
+    pl.col('date') == pl.col('date').max())
+
+vac_min = pl.concat([date_eos.select('cumulative'),
```
swo, line 103: do these get used for anything? vac_min and vac_max

scripts/evaluation_script.py
swo, line 1: There should be a statement (eg in the README) about what this file does, if there is data required for it to run, etc.

swo (changes requested):
It's good to get some of this code in. We'll want a test suite and a consistent interface for testing functions, but we could leave those for future PRs

iup/eval_projections.py
Fuhan-Yang: Yes. The rows that has the mutual `date` between `data_df` and `pred_df` will be selected.

iup/eval_projections.py
Fuhan-Yang: Right, this function just pulls out the total uptake at the end of the season. I can make this function to calculate the difference in the total uptake between the predicted and the observed data, so it will have the same signature.

scripts/evaluation_script.py
Fuhan-Yang: This is the getting the start of the forecast date. So we will know how the predicted end-of-season uptake changes over time.

iup/eval_projections.py
swo: right, like I think you want this to be something like absolute error at end of season (percentage points? or % error?)

scripts/evaluation_script.py
swo: oh, I see, so the `i` is an index over forecasting dates?

I think it would be easier to understand if the models produce data frames with columns like forecasting date, observation date, and forecasted estimate. Because right now the starting forecast date isn't explicit, it's just implicit in being the first row of the data frame of predictions

scripts/evaluation_script.py
Fuhan-Yang: Addressed in [5f2e366](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/37/commits/5f2e366c5115d1e769b602f4aacb4d854cce534e)

scripts/evaluation_script.py
Fuhan-Yang: Addressed in [5f2e366](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/37/commits/5f2e366c5115d1e769b602f4aacb4d854cce534e)

scripts/evaluation_script.py
Fuhan-Yang: Sorry! These didn't get used. Already deleted. See [5f2e366](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/37/commits/5f2e366c5115d1e769b602f4aacb4d854cce534e)

iup/eval_projections.py
Fuhan-Yang: Sounds good. In next PR incorporating the OOP, I will add the absolute error relative to the observed end-of-season uptake: like abs(data-predicted)/data to see how deviated the prediction is compared to the data.

scripts/evaluation_script.py
Fuhan-Yang: Sounds good. I will add the statements in the OOP PR.

Fuhan-Yang:
> It's good to get some of this code in. We'll want a test suite and a consistent interface for testing functions, but we could leave those for future PRs

In the next OOP PR, I plan to add the "bias" metric in addition to MSPE and absolute error of end-of-season uptake. The metrics that relate to the interval estimate will be added once the intervals are available.

PR #41, opened 2024-11-01
Use config template

swo:
- Make the existing `config.yaml` into a template
- In the README, explain that you need to make your own config
- `scripts/main.py` looks for this config as a command line argument
- Auto-formatting in files that have not yet gone through pre-commit compliance


eschrom (approved):
Looks good to me. I will merge and give it a go.

PR #52, opened 2024-11-08
Evaluation architecture

Fuhan-Yang:
This is my preliminary trial on building evaluation architecture. It follows the diagram below, where the blue is included in the code, the grey is saved for future use. Feel free to leave any comments, thoughts, and ideas!
Picture1


iup/eval_objects.py
```diff
@@ -0,0 +1,268 @@
...
+        Maybe can belong to TimelessPointMetric because not every date needs to be matched,
+        but the situation when the last date does not match needs to be defined.
+        -------------------
+        Input: self (joined data)
+        Return: pl.DataFrame with absolute error in the total uptake between data and prediction
+                and the forecast end date.
+        """
+        joined = self.with_columns(
+                cumu_data = pl.col('estimate').cum_sum(),
+                cumu_pred = pl.col('estimate_right').cum_sum()
```
eschrom, line 230: Looks like this is assuming that the input PointForecast objects `data` and `pred` are derived from type IncidentUptakeData. If they were derived from CumulativeUptakeData, the `.cum_sum()` would be unnecessary. There should be some way to check or enforce which UptakeData type is used to make a PointForecast object.

iup/eval_projections.py
```diff
@@ -1,5 +1,100 @@
...
+        pred_df,
+        var
+):
+    """
+    Calculate the mean squared prediction error between the observed
+    data and prediction when both are available.
+
+    Parameters:
+    --------------
+    data_df: polars.DataFrame
```
eschrom, line 16: Perhaps we should insist that `data_df` and `pred_df` are IncidentUptakeData polars dataframes, specifically.

iup/eval_projections.py
```diff
@@ -1,5 +1,100 @@
...
+
+    if not any(data_df['date'].is_in(pred_df['date'])):
+        ValueError('No matched dates between data_df and pred_df.')
+
+    common_dates = data_df.filter(
+        pl.col('date').is_in(pred_df['date'])
+    ).select('date')
+
+    if len(common_dates)!= common_dates.n_unique():
+        ValueError('Duplicated dates are found in data_df or pred_df.')
```
eschrom, line 40: These same checks are performed in the validation of a TimewisePointMetric, so it seems like we should just initialize a TimewisePointMetric here, rather than explicitly duplicating the validation steps.

iup/eval_projections.py
```diff
@@ -1,5 +1,100 @@
...
+            how = 'inner'
+        ).with_columns(
+            spe = (pl.col(var) - pl.col(f"{var}_right"))**2
+        ).with_columns(
+            mspe = pl.col('spe').mean(),
+        ).filter(
+            pl.col('date') == pl.col('date').min(),
+        ).select(
+            'date','mspe'
+        )
```
eschrom, line 55: Similarly, these steps are performed by the `get_mspe` method, above. So we should call that method here, rather than explicitly repeating it as a new function.

iup/eval_projections.py
```diff
@@ -1,5 +1,100 @@
...
+
+
+    return mspe
+
+
+# get end-of-season total uptake #
+
+def get_eos(
```
eschrom, line 63: Same comment as above - `get_eos` is already a method of a TimewisePointMetric, so that method should be called as part of evaluating projections, rather than making an essentially identical standalone function.

scripts/evaluation_script.py
```diff
@@ -48,3 +53,55 @@
     pic_loc="output/",
     pic_name="weekly_predicted_uptake.png",
 )
+
```
eschrom, line 56: I think it would be really useful to add this evaluation and plotting script to the end of main.py in the /scripts folder, rather than writing a separate script. main.py already produces an LIUM with both incident and cumulative projection attributes, plus `split_train_test` can be used to make your test data. So calculating and plotting evaluation metrics could be a natural extension of that script. More importantly, it would force you to use the output of my code as the input to yours, so that we can discover any issues at this interface.

eschrom (commented):
Nice work, Fuhan!! I've left several comments which may be important to address before merging. The last one, in particular, is important - we should make an effort to use the output of my code as the input to yours, to help us discover any compatibility issues. Many of the other comments are aimed at this, too, as well as avoiding repetition, etc. See what you think and let me know if I'm wrong about anything!

iup/eval_objects.py
```diff
@@ -0,0 +1,268 @@
...
+    """
+
+    def __init__(self):
+
+        super().__init__()
+        self.validate()
+
+    # Must have response variable and date #
+    # TODO: make the response variable to be a variable not hardcoded #
+    def validate(self):
```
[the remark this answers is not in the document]
eschrom: These are now their own methods, but we have them defined separately under PointForecast and UptakeData objects. Perhaps something to consolidate in the future.

scripts/evaluation_script.py
eschrom: Or, as Scott suggested, we'll use a makefile to essentially execute our scripts sequentially. Even in this case, however, `get_uptake_data` doesn't exist as a function anymore. It has been superseded by `parse_nis` which can be given its arguments from a .yaml config.

swo:
I would reframe the flow chart above as saying:

- There is a `Forecast` class, which is an extension of a polars dataframe that enforces some schema (say, it has columns `forecast_date`, `observation_date`, `target`, `value`)
- There are two subclasses to `Forecast`: `QuantileForecast` and `SampleForecast`. The quantile forecast will have a column `quantile` (with values like `0.50`) and the sample forecast will have a column `sample_id` (running from 1 to the number of posterior samples).
- `PointForecast` could be a subclass of `QuantileForecast` that enforces there being only a single quantile (ie `0.50`, the median). Or it could be its own subclass of `Forecast` that doesn't have `quantile` or `sample_id`.

I don't know that you need the time/less distinction?

A model like the one that's been coded up now would produce `PointForecast` for now, and could be upgraded to do `QuantileForecast`

iup/eval_objects.py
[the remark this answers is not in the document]
swo: @eschrom See my note above; it looks like there should be a more general `Data` class, and then `UptakeData` and `Forecast` are subclasses of taht.

iup/eval_objects.py
```diff
@@ -0,0 +1,268 @@
...
+    def preprocess(self, data, pred):
+        """
+        Join data and prediction with 1:1 validate
+        """
+        return data.join(
+            pred,
+            on = 'date',
+            how = 'inner',
+            validate = '1:1'
+        )
```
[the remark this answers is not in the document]
swo: `.join()` will rename according to some system (adding `_right` or `_left` as a suffix, I think). Agree that it's better to be explicit and choose the names ourselves

iup/eval_objects.py
swo: Certain metrics will only make sense on certain kinds of forecasts

That could either be with checking the classes of the inputs, or by looking at the `target` column of a `Forecast` object, and making sure that it is, say `"incident_uptake"`

iup/eval_projections.py
```diff
@@ -1,5 +1,100 @@
+import polars as pl
 import altair as alt
 
+# get mspe #
+def get_mspe(
```
swo, line 5: This function is duplicated across these two files? It should be in only one place

swo (changes requested):
High-level comments:

- Let's make a more general `Data` abstract class that can embrace both `UptakeData` and `Forecast`
- The evaluation metrics should be functions (either standalone in a module, or static methods in a class) that take two data frames and return a number. 
- I don't know that you need the `Metric` classes, because there is no relevant state to be tracked there

Happy as always to open suggestion PRs or to talk through synchronously; let me know how I can be helpful.

Fuhan-Yang:
The updated evaluation architecture is that ValidateData as the super parent, embracing UptakeData and Forecast. Forecast has QuantileForecast and SampleForecast, and PointForecast is a subclass of QuantileForecast. All the metrics are standalone functions given the data and the pred are in the matching and standard format.

iup/__init__.py
```diff
@@ -5,9 +5,24 @@
 import abc
 from typing_extensions import Self
 from typing import List
+import re
 
 
-class UptakeData(pl.DataFrame, metaclass=abc.ABCMeta):
+class ValidateData(pl.DataFrame, metaclass=abc.ABCMeta):
```
eschrom, line 11: Do we really need separate classes for observed data and forecasts? In essence, they are both data frames with dates and uptake values. The only difference is in how we interpret them, not in their structure or contents. Maybe we can collapse some of this code...

iup/__init__.py
```diff
@@ -831,3 +846,446 @@ def predict(
...
+        super().__init__(*args, **kwargs)
+        self.validate()
+
+    # Can add validate to enforce the specific schema of Forecast when more Forecast types are added
+    def validate(self):
+        self.assert_type_included(pl.Date)
+        self.assert_type_included(pl.Float64)
+        # has at least 1 column of date and 1 column of estimate
+
+    def assert_type_included(self, datatype):
```
eschrom, line 867: Why not use the `assert_columns_type` function I have already written? It seems to do the same thing.

iup/__init__.py
```diff
@@ -831,3 +846,446 @@ def predict(
...
+        Parameters
+        estimate:
+            The data without 'date' type
+        column_name:
+            The common column name
+        """
+        pattern = rf"^{column_name}\d(\.\d+)?$"
+        assert all(
+            [bool(re.match(pattern, col)) for col in estimate.columns]
+        ), f"Not all columns are Column name {column_name}"
```
eschrom, line 911: Regular expressions always look complicated to me. I bet there is a simpler-looking way to do this, e.g. 
`assert(all(column_name in col for col in estimate.columns))`

iup/__init__.py
```diff
@@ -831,3 +846,446 @@ def predict(
...
+        dtype:
+            data type for each listed column
+        """
+        assert isinstance(columns, List), f"{columns} must be a list"
+
+        for col in columns:
+            assert (
+                self[col].dtype == dtype
+            ), f"Column {col} should be {dtype} but is {self[col].dtype}"
+
```
eschrom, line 958: `assert_columns_found` and `assert_columns_type` are already methods of UptakeData. Instead of repeating these methods here, maybe we should just make them methods of the parent class common to UptakeData and PointForecast.

iup/__init__.py
```diff
@@ -831,3 +846,446 @@ def predict(
...
+    assert any(
+        data["date"].is_in(pred["date"])
+    ), "No matched dates between data and prediction."
+
+    # 2. There should not be any duplicated date in either data or prediction.
+    common_dates = data.filter(pl.col("date").is_in(pred["date"])).select("date")
+
+    assert (
+        len(common_dates) == common_dates.n_unique()
+    ), "Duplicated dates are found in data or prediction."
```
eschrom, line 1062: These checks for matching dates and duplicated dates are performed for several of your evaluation metrics, so maybe they should be their own functions.

iup/__init__.py
```diff
@@ -831,3 +846,446 @@ def predict(
...
+#     """
+#     Evaluation metric, varied by metric type.
+#     """
+
+#     @abc.abstractmethod
+#     def get_metric(self, metric_type):
+#         pass
+
+
+# # Any metric that does not require time-wise matching, can go here.
```
eschrom, line 1133: I haven't read through anything below this point thoroughly, since it is all commented out. Does that mean you intend to save it for future reference? If so, I recommend you delete it here and stash it elsewhere, to try to keep __init__.py as clean as possible.

eschrom (approved):
My comments focus on making the code more concise, largely by integrating our work more tightly. That said, I would be happy to do this for us - it might benefit my understanding of your evaluation code. I would just need you to explain to me the different forecast types you anticipate - their column names, values, etc.

swo (commented):
Before I review, it will help a lot if you respond to the comments I made in the last review. (And if you resolved whatever the concern was, then say that and resolve the conversation.)

This was I know I don't need to worry about repeating myself ;)

swo (commented):
Sorry, this should have been a "comment"

iup/eval_projections.py
Fuhan-Yang: See [issue #57](https://github.com/CDCgov/cfa-immunization-uptake-projection/issues/57)

iup/__init__.py
Fuhan-Yang: Maybe this will be useful when we produce the forecasts that do not have the same format as observed data? Like in sample forecast, the forecast will be a distribution instead of a single Series.

iup/eval_objects.py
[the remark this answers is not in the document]
swo: > See [ad99054](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/52/commits/ad990548dddfb95a284a262cca79330878c28f4f)

It looks like this is just whitespace formatting?

iup/eval_objects.py
```diff
@@ -0,0 +1,288 @@
...
+
+
+class PointForecast(Forecast):
+    """
+    Class for forecast with point estimate
+    Can be a special case for QuantileForecast.
+
+    """
+
+    def __init__(self):
```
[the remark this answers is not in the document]
swo: When you define `validate()` in `PointForecast`, it overrides the definition of `validate()` that was inherited from `Forecast`. So `super().__init__()` calls the same `.validate()` as the line below this does

Prove it to yourself! See this example: https://gist.github.com/swo/4c641bdfbd9d805d42542d7c7290fd0b

iup/__init__.py
Fuhan-Yang: This function is used to match the first common word in the column name, it comes in handy in using like line 895. Or we can use [str.find](https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.str.find.html), which may have better interface.

iup/__init__.py
Fuhan-Yang: Actually I thought about this, but it seems like the methods can only be directly-inherited, i.e, the methods can be passed to its direct subclass only. Or it will be concise to put these two methods in `ValidateData`.

iup/__init__.py
Fuhan-Yang: In this case, we need to put the metrics into a class. @swo what do you think?

iup/__init__.py
Fuhan-Yang: Sure! Sorry for the redundancy.

Fuhan-Yang: test

iup/eval_objects.py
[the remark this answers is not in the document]
Fuhan-Yang: Because I have `self.validate()` followed by `super().__init__()` for each class, this means it is the `validate()` in this class gets called instead of its direct parent class. This will result in that the `validate()` in the ancestor class (here is `ValidateData`) never gets called, but it is fine because it is an abstract class and doesn't have defined `validate()`. Is that correct? The correct way is to delete `self.validate()` for each class.

iup/__init__.py
eschrom: In the case of a sample forecast, the distribution will be given by many (say 1000) samples drawn from it, right? So this is a data frame with a date column and 1000 uptake columns. It still seems to me like this structurally similar enough to be the same class as the observed uptake data, if we design our tests and validate methods right. But maybe I should think about this a little more.

iup/__init__.py
eschrom: If we have 1000 columns for 1000 draws from the posterior, then these columns ought to have well-specified names (e.g. "estimate_1", "estimate_2", etc.). So we could just construct a list of these column names to give to `assert_columns_type`. But again, maybe this is something I will address in the future.

iup/__init__.py
eschrom: Yeah, I think this is another facet of the consolidation theme that runs through most of my comments. We'll decide on a broader strategy first, which will hopefully dictate what we do in this particular case.

eschrom (approved):
I am happy for these changes to be merged; then I volunteer to take on a consolidation effort.

iup/eval_objects.py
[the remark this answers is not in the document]
swo: See #59 and #60

swo (approved):

PR #53, opened 2024-11-08
Generalization for arbitrary grouping factors

eschrom:
This uses the .over() approach to accommodate arbitrary grouping factors. The user now specifies in the config .yaml file which grouping factors to use from each data set, giving both the names of the relevant NIS columns as well as the desired names for these columns. The desired names for the grouping factor columns ought to match across any data sets that will be pooled together into the same model. A new function `extract_group_names` verifies this.

This PR does NOT include any testing - I will save that for later. 

So, please ignore PRs #49, #50, and #51. These should be kept a little while longer for reference, but not merged. This PR is the one that should be thoroughly reviewed and merged to permit arbitrary grouping factors.

### DIFF
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -486,7 +486,7 @@ def extract_group_names(
         [g.get(v) == group_cols[0].get(v) for g, v in zip(group_cols, group_cols[0])]
     )
 
-    group_names = [v for v in group_cols[0].values()]
+    group_names = tuple([v for v in group_cols[0].values()])
 
     return group_names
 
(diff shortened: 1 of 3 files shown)


iup/__init__.py
```diff
@@ -18,15 +18,15 @@ def __init__(self, *args, **kwargs):
...
         self.validate()
+        self.sort("date")
 
     def validate(self):
         """
-        Validate that an UptakeData object has the three key columns:
-        date, grouping factors, and uptake estimate (% of population).
+        Validate that an UptakeData object has the two key columns:
+        date and uptake estimate (% of population). There may be others.
         """
```
Fuhan-Yang, line 27: Since there are multiple functions on `IncidentUptakeData` that require the argument of `group_cols`, maybe it will be helpful to validate the `group_cols` too? As their column names change, maybe we can check that the data should be more than 2 columns (besides `date` and `estimate`)? Just a thought!

Fuhan-Yang (approved):
By generalizing the `regions`, this package is more flexible to take the data with varying grouping factors. This looks good to me!

iup/__init__.py
eschrom: Good catch! We will likely want to accommodate scenarios with no grouping factors, so I will update the code to recognize "None" as a valid entry for group_cols.

PR #62, opened 2024-11-25
Wrapper function for metric calculation

Fuhan-Yang:
This PR creates wrapper functions to include date-match validation and metric calculation in the standalone metric functions.

### DIFF
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -1162,7 +1162,7 @@ def check_date_match(data: IncidentUptakeData, pred: PointForecast):
     ), "Duplicated dates are found in data and prediction."
 
 
-def get_mspe(data: IncidentUptakeData, pred: PointForecast) -> pl.DataFrame:
+def calculate_mspe(data: IncidentUptakeData, pred: PointForecast) -> pl.DataFrame:
     """
     Calculate MSPE between observed data and forecast.
     ----------------------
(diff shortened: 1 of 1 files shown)


iup/__init__.py
```diff
@@ -1132,26 +1133,51 @@ def assert_column_name_all(estimate, column_name):
...
+    pred:
+        The forecast made by model. Should be PointForecast
+
+    Return
+    pl.DataFrame with MSPE and the forecast start date and end date
+
+    """
+    check_date_match(data, pred)
 
     joined = data.join(pred, on="date", how="inner", validate="1:1")
```
eschrom, line 1182: Check with @swo but I think he and I were both suggesting that this calculation of mspe should also be its own function, perhaps called `calculate_mpse`. In other words, `get_mspe` just calls two functions, `check_date_match` and `calculate_mspe` in that order. The advantage is that you can test `calculate_mspe` completely independent of testing `check_date_match` (whereas here you must `check_date_match` in order to test that mspe is calculated correctly).

iup/__init__.py
```diff
@@ -1187,22 +1213,19 @@ def get_mean_bias(data: IncidentUptakeData, pred: PointForecast) -> pl.DataFrame
...
+    pl.DataFrame with mean bias and the forecast start date and end date
+    """
 
-    assert (
-        len(common_dates) == common_dates.n_unique()
-    ), "Duplicated dates are found in data or prediction."
+    # Check the conditions for date match:
+    check_date_match(data, pred)
 
     joined = (
```
eschrom, line 1230: Same comment as above for mspe.

iup/__init__.py
```diff
@@ -1227,10 +1250,22 @@ def get_eos_abe(data: IncidentUptakeData, pred: PointForecast) -> pl.DataFrame:
...
+    Return
+    pl.DataFrame with absolute error in the total uptake between data and prediction
             and the forecast end date.
+
     """
+
+    # Check the conditions for date match:
+    check_date_match(data, pred)
+
     joined = (
```
eschrom, line 1269: Same comment as above for mspe and mean bias.

eschrom (commented):
See the one detailed comment - I think functions for just the metric calculations that do not include checking for matching dates would be helpful for testing purposes. But it's up to you whether you address this now or when you write tests.

PR #64, opened 2024-11-26
Refactor data type and forecast type

Fuhan-Yang:
module_component
 
This PR deletes the intermediate classes between ValidatedData and the end data types for observed data (Incident/CumulativeUptakeData) and forecasts (Quantile/SampleForecast). Each class has slightly different validation methods. In this way, the ValidatedData is a pool of functions, where different functions are used by multiple subclasses. The pro of this architecture is it is clear and easy to understand, while the con is that some functions are not meaningful for some classes but they can still be called. This is explicit for `split_train_test` function, which should not be available for any forecast types. Maybe consider to put it under both Incident&CumulativeUptakeData. However, other functions like 'date_to_*' can be useful for forecast types as well. 

Another note is that, the outcome from `predict()` in the LinearIncidentUptakeModel should be PointForecast type to feed into metric function.

The UptakeData is replaced with ValidatedData in `test_uptake_data.py`, please correct if anything is done wrong!

### DIFF
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -8,33 +8,18 @@ from typing import List
 import re
 
 
-class ValidateData(pl.DataFrame, metaclass=abc.ABCMeta):
+class ValidatedData(pl.DataFrame):
     """
     Abstract class for observed data and forecast data.
     """
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
-
-    @abc.abstractmethod
-    def validate(self) -> None:
-        pass
-
-
-class UptakeData(ValidateData):
-    """
-    Abstract class for different forms of uptake data.
-    """
-
-    def __init__(self, *args, **kwargs):
-        """
-        Initialize an UptakeData object as a polars data frame plus validation.
-        """
-        super().__init__(*args, **kwargs)
+        self.validate()
 
     def validate(self):
         """
-        Validate that an UptakeData object has the two key columns:
+        Validate that an ValidatedData object has the two key columns:
         date and uptake estimate (% of population). There may be others.
         """
 
@@ -68,6 +53,20 @@ class UptakeData(ValidateData):
                 self[col].dtype == dtype
             ), f"Column {col} should be {dtype} but is {self[col].dtype}"
 
+    def assert_type_included(self, datatype):
+        """
+        Verify at least one column has the expected type.
+
+        Parameters
+        columns: List[str]
+            names of columns for which to check type
+        datatype:
+            data type for each listed column
+        """
+        assert any(
+            dtype == datatype for dtype in self.schema.values()
+        ), f"No column is {datatype} type"
+
     def with_columns(self, *args, **kwargs):
         """
         Copy of polars with_columns that returns the same subclass as it's given.
@@ -75,6 +74,22 @@ class UptakeData(ValidateData):
         orig_class = type(self)
         result = super().with_columns(*args, **kwargs)
         return orig_class(result)
+
+    @staticmethod
+    def assert_column_name_all(estimate, column_name):
+        """
+        Verify that all columns have a pattern with a common name
+
+        Parameters
+        estimate:
+            The data without 'date' type
+        column_name:
+            The common column name
+        """
+        pattern = rf"^{column_name}\d+$"
+        assert all(
+            [bool(re.match(pattern, col)) for col in estimate.columns]
+        ), f"Not all columns are Column name {column_name}"
 
     @staticmethod
     def date_to_season(date_col: pl.Expr) -> pl.Expr:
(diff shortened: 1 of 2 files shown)


iup/__init__.py
```diff
@@ -8,33 +8,18 @@
 import re
 
 
-class ValidateData(pl.DataFrame, metaclass=abc.ABCMeta):
+class ValidatedData(pl.DataFrame):
```
eschrom, line 11: Although it doesn't affect how the code works, "ValidatedData" is a weird name to me. What do you think about changing the name of the parent class to "Uptake" and the names of the child classes to "IncidentData," "CumulativeData," "QuantileForecast," "PointForecast," and "SampleForecast" (the last three already being correct)? This makes more sense to me - it's all uptake, but the distinction is between data vs. forecasts. But of course I'm happy to do this in a separate PR, since it's just a lot of nuisance work!

iup/__init__.py
```diff
@@ -76,6 +75,22 @@ def with_columns(self, *args, **kwargs):
         result = super().with_columns(*args, **kwargs)
         return orig_class(result)
 
+    @staticmethod
```
eschrom, line 78: It would be nice to have all these `assert_something` methods as consistent as possible - at least all static methods or all normal methods. Could `assert_column_name_all` take `self` as its first argument and remove the "date" column internally?

iup/__init__.py
```diff
@@ -76,6 +75,22 @@ def with_columns(self, *args, **kwargs):
...
+    def assert_column_name_all(estimate, column_name):
+        """
+        Verify that all columns have a pattern with a common name
```
eschrom, line 81: What if there are grouping factors (e.g. geographic area, demographic group, etc.)? In this case, even a forecast would have columns that are not "date" but also don't contain the "column_name."

iup/__init__.py
```diff
@@ -996,112 +1010,42 @@ def predict(
...
-    Abstract class for all the forecast data type.
+    Class for forecast with quantiles.
+    Save for future.
     """
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
-    # Can add validate to enforce the specific schema of Forecast when more Forecast types are added
+    # Must be named as "quantileXX" except the date column
```
eschrom, line 1022: As noted above, columns for grouping factors seem necessary but don't fit this naming system.

iup/__init__.py
```diff
@@ -996,112 +1010,42 @@ def predict(
...
+    For now, enforce the "quantile50" to be "estimate"
     """
 
     def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
 
     def validate(self):
-        self.assert_columns_found(["estimate"])
-        self.assert_columns_type(["estimate"], pl.Float64)
+        super().assert_columns_found(["estimate"])
```
eschrom, line 1044: Having a bank of `assert_something` methods at the level of the parent class and then choosing the right combination of these methods to validate each child class seems clever to me. Nice work, Fuhan!

iup/__init__.py
```diff
@@ -1111,24 +1055,13 @@ def __init__(self, *args, **kwargs):
...
-        estimate = self.select(pl.all().exclude(pl.Date))
-        SampleForecast.assert_column_name_all(estimate, "sample_id")
-
-    @staticmethod
-    def assert_column_name_all(estimate, column_name):
-        """
-        Verify that all columns have a pattern with a common name
+        # has at least 1 column of date and 1 column of estimate
+        self.assert_type_included(pl.Date)
+        self.assert_type_included(pl.Float64)
```
eschrom, line 1060: I still think it would be better to check that the specific columns for the posterior samples are all type Float64, even if there are a lot of these columns. As long as the columns are named consistently (e.g. "Estimate_Sample_0001", "Estimate_Sample_0002", ... "Estimate_Sample_1000"), you could check that all columns beginning with "Estimate_Sample_" have type Float64.

eschrom (commented):
Some of my comments just reflect personal preferences, but the comments saying, "what if there are grouping factors?" might point out a real problem. Please take a look and let me know if I'm missing something!

iup/__init__.py
Fuhan-Yang: "ValidatedData" indicates that all the classes are the data that is validated, where the data means generic data set. While differing 'data' from 'forecast' indicates it is observed data. This makes sense too. I will change to "ValidatedUptake", emphasizing all the classes have been validated when initialized.

Fuhan-Yang: Changed in 4e77134bb18944b64e1e91c9f6b544132bae8c73

iup/__init__.py
Fuhan-Yang: Good point! Changed in 6f6f54b21f3c366380898cf70c259e3e0e389c76

swo (approved):
Let's get this merged in, although I have a few more tweaks to make.

First set of those in is #65

iup/__init__.py
Fuhan-Yang: Great point! See change in 8ead6e304c8f18304e32fbaa24b7f5b216ea247c

iup/__init__.py
Fuhan-Yang: Just added two assert functions that only check on numerical columns, then adding grouping factors won't cause issues. See 8ead6e304c8f18304e32fbaa24b7f5b216ea247c

iup/__init__.py
Fuhan-Yang: See the two new assert functions (8ead6e304c8f18304e32fbaa24b7f5b216ea247c), that check the name and type for each numerical column

Issue #61, opened 2024-11-25
Write a flowchart of computation


swo:
- Inputs (and their types)
- Where do validation happen
- Which scripts are involved, when
- Add validations where needed (cf https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/65#discussion_r1869780144)

Cf. https://github.com/cdcent/fall-virus-model?tab=readme-ov-file#file-and-script-organization for an example

swo:
- The big question is whether you want a single `main.py` to do everything, or if you want to break this up into different pieces. I guess it's fine to start with everything all in once place, and then start splitting it up, as you find that it slows you down, to need to re-run everything. Do you have a sense of how long it takes `main.py` to run now?
- You can drop everything that's inside of `nisapi`; you can just start from `nisapi.get_nis()`
- I find it hard to read some of these colors: ![image](https://github.com/user-attachments/assets/686ca37e-ac4f-4e7f-acc1-e8e9ebb85ef4)

eschrom:
My vote is to keep the entire pipeline in `main.py` for the time being. Before PR #82, `main.py` takes < 1 second. That said, several upcoming changes will cause this to increase:

- Replacing scikit-learn models that generate point estimates with numpyro models that generate full posteriors
- Using multiple models rather than just the one
- Refitting the model each time the forecast date changes, to include all available data, as discussed in PR #82 

If/when these changes make the run-time of `main.py` intolerably long, that's when I suggest we switch to a Makefile to compartmentalize the steps. But personally, for the sake of scientific progress, I'd prefer to focus on these three items themselves before refactoring our scripts.

eschrom:
As for the flow diagram, I agree with @swo that the granular steps inside NIS_API can be left out, and that the low contrast between font color and compartment color is hard to read. Additionally, removing the "projection" vs. "evaluation" to make [model fitting -> making projections -> evaluation metrics] a single linear process, as discussed in PR #82, will change the structure of the flow diagram a bit. 

But I think this is a very good start, and I like your color-coding of data objects vs. key operations vs. validation vs. etc. Maybe you can teach me how to make these flowcharts sometime?

swo:
> My vote is to keep the entire pipeline in `main.py` for the time being. Before PR #82, `main.py` takes < 1 second

[...]

PR #65, opened 2024-12-03
Data refactor

swo:
- Rename the underlying data/forecast object as `Data`
- Simplify schema validations into a single method
- Add a validation that cumulative uptake estimates must be in [0, 1]
- Tweak tests to account for this new validation
- Methods on `Data` and its subclasses return normal `pl.DataFrame`s. Tweak one or two parts of the current uptake model to account for this.

### DIFF
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -978,11 +978,13 @@ class QuantileForecast(Data):
 
     def validate(self):
         self.assert_in_schema(
-            [("date", pl.Date), ("quantile", pl.Float64), ("estimate", pl.Float64)]
+            {"date": pl.Date, "quantile": pl.Float64, "estimate": pl.Float64}
         )
 
         # all quantiles should be between 0 and 1
-        assert self["quantile"].is_between(0.0, 1.0).all()
+        assert (
+            self["quantile"].is_between(0.0, 1.0).all()
+        ), "quantiles must be between 0 and 1"
 
 
 class PointForecast(QuantileForecast):
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -28,11 +28,15 @@ class Data(pl.DataFrame):
         for name, type_ in names_types.items():
             if name not in self.schema.names():
                 raise RuntimeError(f"Column '{name}' not found")
-            elif (name, type_) not in self.schema.items():
+            elif (
+                name in self.schema.names() and (name, type_) not in self.schema.items()
+            ):
                 actual_type = self.schema.to_python()[name]
-                f"Column '{name}' has type {actual_type}, not {type_}"
-
-            assert (name, type_) in self.schema.items()
+                raise RuntimeError(
+                    f"Column '{name}' has type {actual_type}, not {type_}"
+                )
+            else:
+                assert (name, type_) in self.schema.items()
 
 
 class UptakeData(Data):
@@ -1009,7 +1013,7 @@ class SampleForecast(Data):
 
     def validate(self):
         self.assert_in_schema(
-            [("date", pl.Date), ("sample_id", pl.Int64), ("estimate", pl.Float64)]
+            {"date": pl.Date, "sample_id": pl.Int64, "estimate": pl.Float64}
         )
 
 
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -26,10 +26,13 @@ class Data(pl.DataFrame):
             names_types (dict[str, pl.DataType]): Column names and types
         """
         for name, type_ in names_types.items():
-            assert (
-                name,
-                type_,
-            ) in self.schema, f"Column '{name}' of type {type_} not found"
+            if name not in self.schema.names():
+                raise RuntimeError(f"Column '{name}' not found")
+            elif (name, type_) not in self.schema.items():
+                actual_type = self.schema.to_python()[name]
+                f"Column '{name}' has type {actual_type}, not {type_}"
+
+            assert (name, type_) in self.schema.items()
 
 
 class UptakeData(Data):
@@ -153,12 +156,6 @@ class IncidentUptakeData(UptakeData):
     """
     Subclass of ValidatedUptake for incident uptake.
     """
-
-    def validate(self):
-        # perform validations for UptakeData
-        super().validate()
-        # and also require that uptake be a proportion
-        assert self["estimate"].is_between(0.0, 1.0).all()
 
     def trim_outlier_intervals(
         self, group_cols: tuple[str,] | None, threshold: float = 1.0
@@ -284,6 +281,14 @@ class CumulativeUptakeData(UptakeData):
     """
     Subclass of ValidatedUptake for cumulative uptake.
     """
+
+    def validate(self):
+        # same validations as UptakeData
+        super().validate()
+        # and also require that uptake be a proportion
+        assert (
+            self["estimate"].is_between(0.0, 1.0).all()
+        ), "cumulative uptake `estimate` must be a proportion"
 
     def to_incident(self, group_cols: tuple[str,] | None) -> IncidentUptakeData:
         """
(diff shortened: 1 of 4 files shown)


iup/__init__.py
```diff
@@ -4,77 +4,45 @@
...
-        dtype:
-            data type for each listed column
-        """
-        for col in columns:
-            assert (
-                self[col].dtype == dtype
-            ), f"Column {col} should be {dtype} but is {self[col].dtype}"
-
-    def with_columns(self, *args, **kwargs):
+class UptakeData(Data):
```
eschrom, line 38: This class is named "UptakeData," but below there are many spots where "ValidatedUptake" is used (often directly replacing "UptakeData." Looks like we must decide which of these terms we like better. I prefer "UptakeData" because "Validated" seems unnecessary - we know we are going to validate whenever we create one of these objects.

iup/__init__.py
```diff
@@ -4,77 +4,45 @@
...
+    def validate(self):
         """
-        Copy of polars with_columns that returns the same subclass as it's given.
+        Validate that an ValidatedUptake object has the two key columns:
```
eschrom, line 41: "ValidatedUptake" should be "UptakeData"

iup/__init__.py
```diff
@@ -145,47 +113,48 @@ def date_to_elapsed(date_col: pl.Expr) -> pl.Expr:
...
+        uptake_data_list: Sequence[Data]
             cumulative or incident uptake data objects, often from different seasons
         start_date: dt.date
             the first date for which projections should be made
         side: str
             whether the "train" or "test" portion of the data is desired
 
         Returns
-        UptakeData
+        ValidatedUptake
```
eschrom, line 131: "ValidatedUptake" should be "UptakeData"

iup/__init__.py
```diff
@@ -145,47 +113,48 @@ def date_to_elapsed(date_col: pl.Expr) -> pl.Expr:
...
+        else:
+            raise RuntimeError(f"Unrecognized side '{side}'")
 
         return out
 
 
 class IncidentUptakeData(UptakeData):
     """
-    Subclass of UptakeData for incident uptake.
+    Subclass of ValidatedUptake for incident uptake.
```
eschrom, line 157: "ValidatedUptake" should be "UptakeData"

iup/__init__.py
```diff
@@ -199,7 +168,7 @@ def trim_outlier_intervals(
           threshold (float): maximum standardized interval between first two dates
 
         Returns
-        IncidentUptakeData
+        IncidentValidatedUptake
```
eschrom, line 171: "IncidentValidatedUptake" should be "IncidentUptakeData" - why has the reverse substitution been made? The latter is defined as a class and the former is not.

iup/__init__.py
```diff
@@ -309,16 +274,22 @@ def to_cumulative(self, group_cols: tuple[str,] | None, last_cumulative=None):
...
 
-        out = CumulativeUptakeData(out)
-
         return out
 
 
 class CumulativeUptakeData(UptakeData):
     """
-    Subclass of UptakeData for cumulative uptake.
+    Subclass of ValidatedUptake for cumulative uptake.
```
eschrom, line 282: "ValidatedUptake" should be "UptakeData"

iup/__init__.py
```diff
@@ -309,16 +274,22 @@ def to_cumulative(self, group_cols: tuple[str,] | None, last_cumulative=None):
...
     """
 
+    def validate(self):
+        # same validations as UptakeData
+        super().validate()
+        # and also require that uptake be a proportion
+        assert (
+            self["estimate"].is_between(0.0, 1.0).all()
+        ), "cumulative uptake `estimate` must be a proportion"
```
eschrom, line 291: Incident uptake is also a proportion, although we accept the fact that it can sometimes be reported as a negative. Should we implement a similar assertion that each "estimate" in an IncidentUptakeData is between -1.0 and 1.0?

eschrom (changes requested):
Looks like there is still some confusion about naming things. For example, as far as I can tell, there is a class called "UptakeData" and no class called "ValidatedUptake," but "ValidatedUptake" keeps showing up throughout the code, often where "UptakeData" should be.

iup/__init__.py
Fuhan-Yang: The `UptakeData` was replaced with `ValidatedData` in PR#64, while this branch is opened before PR#64 merged to main, the original `UptakeData` is reserved. Then `ValidatedData` is changed to `ValidatedUptake` based on this [comment](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/64#discussion_r1867830878).

iup/__init__.py
Fuhan-Yang: This is an error in replacing, it should be `IncidentUptakeData`!

iup/__init__.py
eschrom: Makes sense as to how we got here! But we still must decide among these names, and my vote still for "UptakeData" as opposed to anything with "Validated" in it.

Fuhan-Yang: I'm fine using `UptakeData` to include observed data (`IncidentUptakeData`) and predictions ('QuantileForecast`) since they are both uptake data set. Let us know your thought! @swo

swo: I just completed a rebase; I called the top level class just `Data`. Open to changes.

iup/__init__.py
swo: The way I would think about this is: I know that cumulative uptake should always be between 0 and 1, and my validation of incident uptake should be that I can convert it to cumulative without error

I'll make this a separate issue

swo: cf #68

eschrom: works for me!

eschrom (approved):
There are still several places (in commented code) that refer to a "ValidatedUptake" class that doesn't exist anymore. Doesn't really matter if that gets fixed now or later.

Fuhan-Yang (approved):
My comments are about small changes to adapt to the new assert_schema method, and a question about if using `validate` when initiating an instance or after. Happy to open another PR.

iup/__init__.py
```diff
@@ -1010,21 +970,19 @@ def predict(
...
-
-    # Must be named as "quantileXX" except the date column
     def validate(self):
-        # must have a date column, and numerical columns must have common name and same data type
-        super().assert_columns_type(["date"], pl.Date)
-        super().assert_numerical_columns_type(pl.Float64)
-        super().assert_numerical_columns_name("quantile")
+        self.assert_in_schema(
+            [("date", pl.Date), ("quantile", pl.Float64), ("estimate", pl.Float64)]
+        )
```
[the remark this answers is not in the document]
swo: I think this is #68 . Incident uptake can be negative, but the corresponding cumulative uptake should not go outside of [0, 1], which is sort of confusing

iup/__init__.py
```diff
@@ -1034,28 +992,23 @@ class PointForecast(QuantileForecast):
...
-    def __init__(self, *args, **kwargs):
-        super().__init__(*args, **kwargs)
-
     def validate(self):
-        super().assert_columns_found(["estimate"])
-        super().assert_columns_type(["estimate"], pl.Float64)
+        # same validations as for QuantileForecast
+        super().validate()
+        # but additionally require that there be only one quantile value
+        assert (self["quantile"] == 0.50).all()
```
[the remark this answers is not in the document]
swo: I'm going to save this for #61

iup/__init__.py
```diff
@@ -1010,21 +970,19 @@ def predict(
...
 
 #### prediction output ####
-class QuantileForecast(ValidatedUptake):
+class QuantileForecast(Data):
     """
     Class for forecast with quantiles.
     Save for future.
     """
 
-    def __init__(self, *args, **kwargs):
```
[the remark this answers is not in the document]
swo: Because `QuantileForecast` is a subclass of `Data`, it inherits any method in `Data`, including `__init__()`, that isn't overwritten

See https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/65/commits/443f26fc326908e6342aa0a98f3ab75146f6f6c9 for a test that shows this works

(but also, I hadn't written a test, so I still had the validation as a list of tuples, rather than a dictionary. Good catch!)

iup/__init__.py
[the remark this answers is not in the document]
Fuhan-Yang: In addition to restraining incident uptake (as in PR #68), the `quantile` column should be checked too because it should always between [0,1]

PR #66, opened 2024-12-03
Reorg modules

swo:
- Separate into `eval` and `models` modules
- Move methods specific to the IncidentUptakeModel stuff into that class
- Add some tests

cc @Fuhan-Yang @eschrom


tests/test_linear_incident_uptake_model.py
```diff
@@ -156,11 +157,146 @@ def test_project_sequentially():
...
         elapsed, start, standards, model
     )
 
     assert all(
         np.round(np.delete(output, 0), decimals=10)
         == np.round(np.array(elapsed), decimals=10)
     )
+
+
+def test_trim_outlier_intervals_handles_two_rows(frame):
```
eschrom, line 170: I gently disagree with the placement of tests for `trim_outlier_intervals`, `augment_implicit_columns`, and `date_to_whatever` in the "test_linear_incident_uptake_model.py" file. I would keep them in "test_uptake_data.py" because they are operations on data that can be performed regardless of which model, if any, is being built. I understand that the linear incident uptake model is the only model that requires these operations _right now_, but what happens if several types of models require these operations as pre-requisites in the future? In that case, it would seem their tests are misplaced.

eschrom (approved):
Overall, I'm fine with however you want to reorganize into submodules. The biggest hurdle is for me to get used to importing & referencing these submodules separately, which is important for me to learn. That said, I've included a minor critique that you can take or leave at your discretion.

tests/test_linear_incident_uptake_model.py
swo: We agree that the structure of the code should match the structure of the tests: if some method is a method of `Data`, it should be tested in or near the place that other `Data` methods are being tested.

I think we're disagreeing about where to put pre-processing functionality that is currently not shared between models but which could become shared. I lean the other way from you because (1) I put a premium the cleanliness of the top-level `__init__.py` namespace and (2) in my experience it's usually more costly to pre-optimize (e.g., by trying to build generic functionalities for input data manipulation) before seeing what the range of model types is.

eschrom: Yep, that's the conflict. Perhaps I'll come around to your way of thinking, so I'm happy to keep it as is.

Fuhan-Yang (approved):
The structure looks good to me. While the script in eval.py has been updated (see `__init__.py` in main), which can be updated now or later.

PR #69, opened 2024-12-03
Refactor wrappers

swo:
- There is one wrapper function `score()` that takes `pred` and `data` frames and a score function
  - This is where the validation happens
  - There is only one test for this function
- The score functions take in two vectors and return a single number
  - Each score function gets a very simple vector test

### DIFF
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -4,7 +4,7 @@ from sklearn.linear_model import LinearRegression
 import numpy as np
 import abc
 from typing_extensions import Self
-from typing import List
+from typing import List, Callable
 import re
 from polars import testing
 
@@ -1088,9 +1088,13 @@ def check_date_match(data: IncidentUptakeData, pred: PointForecast):
     ), "Duplicated dates are found in data and prediction."
 
 
-def calculate_mspe(data: IncidentUptakeData, pred: PointForecast) -> pl.DataFrame:
+def score(
+    data: IncidentUptakeData,
+    pred: PointForecast,
+    score_fun: Callable[[pl.Expr, pl.Expr], pl.Expr],
+) -> pl.DataFrame:
     """
-    Calculate MSPE between observed data and forecast.
+    Calculate score between observed data and forecast.
     ----------------------
 
     Parameters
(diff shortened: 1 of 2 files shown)


iup/__init__.py
```diff
@@ -1088,186 +1088,64 @@ def check_date_match(data: IncidentUptakeData, pred: PointForecast):
...
+        Scoring function. Takes observed and true values.
 
     Return
-    pl.DataFrame with forecast start date, forecast end date, and MSPE
+    pl.DataFrame with one row: forecast start date, forecast end date, and score
 
     """
-    joined = data.join(pred, on="date", how="inner", validate="1:1")
+    # validate inputs
+    data = IncidentUptakeData(data)
```
eschrom, line 1113: Our evaluation metrics assume that the data and predictions are incident (not cumulative) uptake. So here, should we _check_ that `data` is `IncidentUptakeData` rather than _coercing_ it into `IncidentUptakeData`? Seems like one could accidentally provide `CumulativeUptakeData` (which passes all the `IncidentUptakeData` validation, I believe) and get a (wrong) answer with no error.

tests/test_evaluation.py
```diff
@@ -35,64 +34,40 @@ def pred():
...
 
-    assert output_date.equals(
-        pl.DataFrame(
-            {"forecast_start": date(2020, 1, 1), "forecast_end": date(2020, 1, 5)}
-        )
-    )
+def test_mspe():
+    x = np.array([0.0, 0.1, 0.7, 0.4, 0.5])
+    y = np.array([0.0, 0.2, 1.0, 0.6, 0.5])
+    assert np.isclose(iup.mspe(x, y), 0.028)
```
eschrom, line 55: Haha `isclose` is new to me. Should be useful in some of the other model-building tests, too. Python has everything.

eschrom (approved):
One minor suggestion, but overall looks like a great way to streamline a little!

iup/__init__.py
swo: fair, maybe this should be `isinstance(data, IncidentUptakeData)`?

eschrom: Yeah, I think that would be a good fix.

Fuhan-Yang: Is the validation already done in the arguments? Like the score function explicitly asks for `IncidenUptakeData` as the type of data.

PR #73, opened 2024-12-06
Refactor PR#67 & PR#69

Fuhan-Yang:
PR#67 was closed after PR#69 merging in, to avoid confusion and conflict. This PR includes all the requested changes from PR#67 and PR#69.

### DIFF
--- a/iup/eval.py
+++ b/iup/eval.py
@@ -1,117 +1,96 @@
-from iup import IncidentUptakeData, PointForecast
 import polars as pl
+from iup import IncidentUptakeData, PointForecast
+from typing import Callable
 
 
-def get_mspe(data: IncidentUptakeData, pred: PointForecast) -> pl.DataFrame:
+###### evaluation metrics #####
+def check_date_match(data: IncidentUptakeData, pred: PointForecast):
     """
-    Calculate MSPE between data and pred
+    Check the dates between data and pred.
+    Dates must be 1-on-1 equal and no duplicate.
     ----------------------
-    Input: data, pred in a matching and validated format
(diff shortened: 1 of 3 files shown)


iup/eval.py
```diff
@@ -1,117 +1,96 @@
...
-    # 2. There should not be any duplicated date in either data or prediction.
-    common_dates = data.filter(pl.col("date").is_in(pred["date"])).select("date")
+    # 1. Dates must be 1-on-1 equal
+    (data["date"] == pred["date"]).all()
 
+    # 2. There should not be any duplicated date in either data or prediction.
     assert (
-        len(common_dates) == common_dates.n_unique()
-    ), "Duplicated dates are found in data or prediction."
+        len(data["date"]) == data["date"].n_unique()
```
swo, line 32: [low priority] Another way to do this is with https://docs.pola.rs/api/python/stable/reference/expressions/api/polars.Expr.is_duplicated.html 

I think it would be:
```python
assert data["date"].is_duplicated().any().not_()
```

tests/test_evaluation.py
```diff
@@ -0,0 +1,75 @@
+import polars as pl
+import numpy as np
+import iup
+from iup import eval
+import pytest
+from datetime import date
+
+
+@pytest.fixture
```
swo, line 9: Because we now only use the `data()` and `pred()` fixtures for one test, you could just move the definition of those data frames into that one test

but not super high priority at all

swo (approved):
Thanks for fixing up our PR mess! Some thoughts but they are just minor improvements, not gating.

PR #74, opened 2024-12-09
Demo numpyro model

### LINKED ISSUE #38: Make a model that outputs posterior samples
See eg https://github.com/cdcent/cfa-covid-booster-uptake-projection/commit/151214044a535517a288eaf83dde90b382a89b54 for a numpyro model

swo:
Resolves #38 

This isn't intended to be a fully-working example, just a way to show how numpyro models are structured

### DIFF
--- a/scripts/demo_numpyro.py
+++ b/scripts/demo_numpyro.py
@@ -22,9 +22,15 @@ def model_factory(
 ):
     """Make a curve fitting model
 
+    Given priors and a parametric curve, return a model, i.e., the kind of function that can be passed to
+    the sampler. Output values from the curve are
+
     Args:
         priors (dict): mapping from variable name to numpyro.distributions object
-        curve (function): (x, pars) -> y
+        curve (function): (x,  sigma, pars) -> y, where:
+          - x and y are arrays of the input and output curves
+          - sigma is the standard deviation of each y value
+          - pars is a dictionary of parameters, from string name to a numpyro distribution
 
     Returns:
         function to be called with MCMC(NUTS(...))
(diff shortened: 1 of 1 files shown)


swo:
@eschrom @Fuhan-Yang Some things to note here:

- The "model factory" setup makes sense if you want the option to slot in different kinds of parametric curves. If it would be useful to see a simpler example, where the curve is hard-coded into the model, happy to show that
- The use of the word "model" here is numpyro jargon. "Model" is the specific thing that gets passed to the MCMC sampler. I'm not *a priori* certain how the posterior-sample-generator "models" that we've been thinking about relate to this kind of object. cc @afmagee42 who might be able to say more on this.

afmagee42:
I think this is on the right track, and in some ways it's a more refined version of what [I was sketching](https://github.com/CDCgov/cfa-immunization-uptake-projection/blob/afm-model-sketch/scratch/models.py). I don't think this is quite right, though, and at a glance it looks to me like this
- Forbids autoregressive-type models when with a little care those are not a problem
- Makes it harder to tack on multiple data streams (what if we have a different observation model for a different survey?)

I think the right paradigm looks more like:
- A model for a latent timeseries is a thing that produces values for some times.
- A likelihood is a thing that knows what the observation model is, and what we've observed

Then we can have a broader conception of models, and we don't ever have to worry about predicting ever again. We just let the model sample the observations for timepoints we don't have data.

afmagee42:
I do like that this splits out functional form from prior distribution choice, which makes comparing prior predictives easier.

swo:
The point here was to demo the simplest kind of numpyro-relevant model I could think of, which is one that has a single uptake curve, with some parameters. This is not intended to be a complete architecture choice, just a way for @eschrom and @Fuhan-Yang to see what the pieces of a numpyro model could look like.

Agree that when we build more complete architecture, we'd probably want it to be able to also do autoregressive models, but I think we should just build some models first, before we get hung up on factories

Like, I built this factory because I wanted to test like 5 different curves and see what fit, so I had good reason to do it

eschrom (commented):
Super helpful - thank you Scott and Andy! I anticipate that, as a learning exercise, I will try to write my own numpyro code "from scratch," but I will end up peeking at this periodically for help. Along the way, I may have questions about this "factory" approach. Maybe we want to convert this PR to a draft and leave it as the forum for such questions?

swo:
Closing now that #102 is open

Issue #40, opened 2024-11-01
Update `Data` objects to use NIS API column name


swo:
NIS API will report start, end, and uptake during that period

So we want `time_end` or similar, rather than `date`

swo:
This is fixed

PR #75, opened 2024-12-09
Use nisapi

swo:
- Add nisapi dependency to lockfile
- Refactor `scripts/main.py` to use `nisapi`
- Add a Makefile with `make cache` and `make run` rules
- Remove old parsing & cleaning functions and tests

Notes on how caching works:

- Rather than make a global user cache, put the cache right in this directory. This will help with Makefile stuff.

Things that are out of scope:

- Updating validation (cf. #40)
  - `make run` fails because `CumulativeUptakeData` validation fails for data coming from `nisapi`
- Figuring out what slice(s) of the NIS data you want


iup/__init__.py
```diff
@@ -1,7 +1,8 @@
-import polars as pl
 import datetime as dt
 from typing import Sequence
 
+import polars as pl
```
eschrom, line 4: Looks like this is just moved a few lines - any reason? Also, beware there will be conflicts here since the last PR #72 merged, so there are some further imports (e.g. `from typing import List` and `from polars.datatypes.classes import DataTypeClass`) necessary.

iup/__init__.py
```diff
@@ -152,153 +153,6 @@ def to_incident(self, group_cols: tuple[str,] | None) -> IncidentUptakeData:
...
-            date=pl.col(date_col).str.to_date(date_format),
-        )
-        .drop_nulls(subset=["estimate"])
-        .sort("date")
-    )
-
-    return frame
-
-
 def insert_rollout(
```
eschrom, line 156: `insert_rollout` had been called from inside `parse_nis` which is now gone. So `insert_rollout`'s job must be done somewhere else now. Maybe as another class method inside LinearIncidentUptakeModel?

scripts/main.py
```diff
@@ -1,11 +1,21 @@
+import argparse
```
eschrom, line 1: Again it looks like some of these imports are reordered, with blank lines added. Any reason?

scripts/main.py
```diff
@@ -1,11 +1,21 @@
...
+    data = nisapi.get_nis(cache)
+
     # List of the cumulative data sets described in the yaml
-    cumulative_data = [iup.parse_nis(**x) for x in config["data"].values()]
+    # note that df.filter(**{"column1": value1, "column2": value2}) is equivalent to
+    # df.filter(pl.col("column1")==value1, pl.col("column2")==value2)
+    cumulative_data = [
+        iup.CumulativeUptakeData(data.filter(**x["filters"]).collect())
+        for x in config["data"].values()
+    ]
```
eschrom, line 18: This performs the filtering of rows, but unnecessary columns are no longer dropped (this had been done by `parse_nis` and `select_columns` which are now gone). For example, "indicator_type" and "indicator" will tag along as columns in the uptake data, even though they always have the same value for every row. I assume this is by design, but just wanted to note it in case you didn't want that.

scripts/config_template.yaml
```diff
@@ -1,17 +1,15 @@
 # The data sets to load and how to interpret them
 data:
   data_set_1:
-    path: PATH_TO_DATA_GOES_HERE
-    estimate_col: estimate
-    date_col: date
     group_cols:
       geography: region
```
eschrom, line 5: Renaming the group columns (e.g. "geography" -> "region" here) is no longer done (had been part of `select_columns` which is now gone). So we should update this part of the config to give the column names for grouping factors, but not new names (i.e. group_cols is just a list, not itself a dictionary). This will precipitate edits to `extract_group_names` in __init__.py also - see below.

scripts/main.py
```diff
@@ -1,11 +1,21 @@
...
 
     # List of grouping factors used in each data set
     grouping_factors = iup.extract_group_names(
```
eschrom, line 21: `extract_group_names` had accepted a list of dictionaries - the how-to-rename-grouping-columns dictionaries from each data set. But this is no longer necessary. `extract_group_names` should now accept a list of lists - the names-of-grouping-columns from each data set.

tests/test_data_cleaning.py
```diff
@@ -1,7 +1,9 @@
-import iup
+import datetime as dt
+
 import polars as pl
 import pytest
-import datetime as dt
+
+import iup
```
eschrom, line 6: Again, there's some reordering and extra blank lines here - any reason?

eschrom (commented):
Warm welcome to the nis_api interface for getting data! That said, there appear to be a few loose ends (e.g. `insert_rollout` no longer has a home, `extract_group_names` is a little outdated). We may want to fix these or else decide they are out of scope before merging.

iup/__init__.py
swo: The rearrangement is related to #76 , which I just opened to formalize this

I'll rebase and catch those other points!

scripts/main.py
swo: See #76

scripts/main.py
swo: confusingly, the different NIS data sets refer to "got a vaccination" in slightly different ways, with different indicators

so I think this will be some work in whatever script does the pipeline, to figure out exactly what kind of data should get passed in

what's the right place to start here? Covid during 2023/2024?

tests/test_data_cleaning.py
swo: See #76

scripts/main.py
swo: I think now we will have the same grouping columns from every dataset, since the data schema is the same. It should always be `geography` and `domain`, right? (This should still be a function argument, but I think we just need a list, because we can assume that all incoming datasets have the same schema.)

The `"filters"` bit I think you still want, because a "dataset" is defined by taking the whole pile of NIS data and subsetting based on vaccine, indicator, etc.

Although I'm now thinking we might want more sophisticated filters, like start and end dates? Or there could be some preprocessor that takes the NIS data, pulls out just the uptake indicator, labels by season, etc. Because I think that's what we're going to want for all the models

iup/models.py
```diff
@@ -610,3 +611,38 @@ def date_to_interval(date_col: pl.Expr) -> pl.Expr:
         Time difference is always in days.
         """
         return date_col.diff().dt.total_days().cast(pl.Float64)
+
+    @staticmethod
+    def insert_rollout(
```
eschrom, line 616: Now `insert_rollout` is in the right place, but it still isn't being deployed. Fixing that seems out-of-scope for this PR though. I filed this into issue #77 and will deal with it later this sprint.

scripts/main.py
eschrom: 2023/2024 covid sounds like a fine starting point. As long as each data set is filtered to only the "got a vaccination" rows (regardless of how it is worded), then the "indicator_type" and "indicator" columns can be dropped altogether. Which can be done in the script.

scripts/main.py
eschrom: Yes, I think you are right that "geography" and "domain" are the only grouping factor column names possible. So all `extract_group_names` must do is determine which/neither/both of these should be used consistently across all data sets being used together. See Issue #78.

I'm not sure we need more sophisticated pre-processing, because the "filters" entry in the config will pull out just the uptake indicator, and `date_to_season` can already label by season.

eschrom (approved):
Looks good! The loose ends are now pushed out-of-scope, so I added issues for them.

scripts/main.py
swo: I started a `scripts/preprocess.py` script and added it to the Makefile. We'll probably want to reconfigure how that reads from the config, etc. But it's a starting point, and the scope of this PR is just to get us started with pulling data from the API, rather than a specific file on disk.

scripts/config_template.yaml
swo: I made a small change, but I'll leave the reworking of the pipeline for another PR

scripts/main.py
swo: I realized that you only need to filter `indicator` for `"received a vaccination"`, which is consistent across datasets. So we can drop `indicator_type` and `indicator`. That's what's in `preprocess.py` for now; I expect we'll refactor a lot of this later.

PR #76, opened 2024-12-11
Update ruff pre-commit hooks

swo:
To harmonize with https://github.com/cdcent/cfa-repo-template/pull/38

The `ruff check` hook does what we used to use [isort](https://pycqa.github.io/isort/) to do

See https://github.com/cdcent/cfa-predict-handbook/pull/339/files for how to get VSCode to do this work automatically on save, so that pre-commit and VSCode are doing the same thing

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -1,27 +1,30 @@
 repos:
-#####
-# Basic file cleanliness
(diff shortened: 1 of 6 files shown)


eschrom (approved):
Admittedly, I don't know how pre-commit works - all I know is that I want it to work! I'll trust you on this and update my VS Code settings as you recommended!

Issue #78, opened 2024-12-11
Refactor `extract_group_names`


eschrom:
NIS API only gives "geography" and "domain" as possible grouping variables, so simplify `extract_group_names` and the config "group_cols" entry to suit this.

eschrom:
Addressed with #79

Issue #77, opened 2024-12-11
`insert_rollout` as part of linear incident uptake model construction


eschrom:
NIS API supersedes most of the data pre-processing that had been done, but `insert_rollout` still must be done to make a linear incident uptake model work correctly.

eschrom:
Addressed with #79

PR #79, opened 2024-12-12
Adjustments to insure compatibility with NIS API

eschrom:
After merging #75, some incompatibilities surfaced:

- NIS API data have different column names than our data objects expect (e.g. "time_end" vs. "date"). Also, NIS API data have many columns that are not grouping factors (e.g. "vaccine", "indicator", "indicator_type"). To solve this, after calling on NIS API, `main.py` now drops and renames columns to achieve the minimalist schema expected by our data objects like CumulativeUptakeData. This extra behavior is controlled from the config file.
- NIS API data may encompass multiple seasons (whereas the vax uptake code had tacitly assumed exactly one season per data set). Relatedly, NIS API data does not include the rollout dates for each season. To solve this, `insert_rollout` can now recognize multiple rollout dates per data set. This behavior is also controlled from the config file. In a change of heart, I have moved `insert_rollout` to be a method of CumulativeUptakeData, because inserting 0.0 cumulative uptake on the rollout date is not necessary for any model, but it could be useful for multiple models. This addresses issue #77. 
- NIS API data have standardized column names, so there is no need to rename the grouping factor columns if not desired. As a result, `extract_group_names` can accept a list of lists of grouping factor columns (one list per data set), rather than a list of dictionaries. To solve this, `extract_group_names` has been refactored accordingly. This addresses issue #78. 
- The pytests in `test_data_cleaning.py` are updated to reflect these changes.

Note: the updated config_template.yaml reflects my usage of 2023-24 covid booster uptake data, ksfb-ug5d, as the input for `main.py`

### DIFF
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -223,7 +223,8 @@ class CumulativeUptakeData(UptakeData):
                 .with_columns(
                     min=(pl.col("estimate") - pl.min("estimate")).over((group_cols))
                 )
-                .filter(pl.col("time_end").first().over(group_cols))
+                .group_by(group_cols)
+                .first()
             )["min"]
             == 0
         ).all()
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -154,14 +154,14 @@ class CumulativeUptakeData(UptakeData):
         return IncidentUptakeData(out)
 
     def insert_rollout(
-        self, rollout: dt.date, group_cols: tuple | None
+        self, rollout: List[dt.date], group_cols: tuple | None
     ) -> pl.DataFrame:
         """
-        Insert into cumulative uptake data rows with 0 uptake on the rollout date.
+        Insert into cumulative uptake data rows with 0 uptake on rollout dates.
 
         Parameters
-        rollout: dt.date
-            rollout date
+        rollout: List[dt.date]
+            list of rollout dates
         group_cols: tuple[str] | None
             names of grouping factor columns
 
@@ -171,16 +171,19 @@ class CumulativeUptakeData(UptakeData):
         Details
         A separate rollout row is added for every grouping factor combination.
         """
-        if group_cols is not None:
-            rollout_rows = (
-                self.select(pl.col(v) for v in group_cols)
-                .unique()
-                .with_columns(date=rollout, estimate=0.0)
-            )
-        else:
-            rollout_rows = pl.DataFrame({"date": rollout, "estimate": 0.0})
+        frame = self
 
-        frame = self.vstack(rollout_rows.select(self.columns)).sort("date")
+        for r in rollout:
+            if group_cols is not None:
+                rollout_rows = (
+                    frame.select(pl.col(v) for v in group_cols)
+                    .unique()
+                    .with_columns(date=r, estimate=0.0)
+                )
+            else:
+                rollout_rows = pl.DataFrame({"date": r, "estimate": 0.0})
+
+            frame = frame.vstack(rollout_rows.select(frame.columns)).sort("date")
 
         return frame
 
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -153,40 +153,36 @@ class CumulativeUptakeData(UptakeData):
 
         return IncidentUptakeData(out)
 
+    def insert_rollout(
+        self, rollout: dt.date, group_cols: tuple | None
+    ) -> pl.DataFrame:
+        """
+        Insert into cumulative uptake data rows with 0 uptake on the rollout date.
 
-def insert_rollout(
-    frame: pl.DataFrame, rollout: dt.date, group_cols: dict | None
-) -> pl.DataFrame:
-    """
-    Insert into NIS uptake data rows with 0 uptake on the rollout date.
+        Parameters
+        rollout: dt.date
+            rollout date
+        group_cols: tuple[str] | None
+            names of grouping factor columns
 
-    Parameters
-    frame: pl.DataFrame
-        NIS data in the midst of parsing
-    rollout: dt.date
-        rollout date
-    group_cols: dict | None
-        dictionary of the NIS columns for the grouping factors
-        keys are the NIS column names and values are the desired column names
+        Returns
+            cumulative uptake data with rollout rows included
 
-    Returns
-        NIS cumulative data with rollout rows included
+        Details
+        A separate rollout row is added for every grouping factor combination.
+        """
+        if group_cols is not None:
+            rollout_rows = (
+                self.select(pl.col(v) for v in group_cols)
+                .unique()
+                .with_columns(date=rollout, estimate=0.0)
+            )
+        else:
+            rollout_rows = pl.DataFrame({"date": rollout, "estimate": 0.0})
 
-    Details
-    A separate rollout row is added for every grouping factor combination.
-    """
-    if group_cols is not None:
-        rollout_rows = (
-            frame.select(pl.col(v) for v in group_cols.values())
-            .unique()
-            .with_columns(date=rollout, estimate=0.0)
-        )
-    else:
-        rollout_rows = pl.DataFrame({"date": rollout, "estimate": 0.0})
+        frame = self.vstack(rollout_rows.select(self.columns)).sort("date")
 
-    frame = frame.vstack(rollout_rows.select(frame.columns)).sort("date")
-
-    return frame
+        return frame
 
 
 def extract_group_names(
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -83,10 +83,36 @@ class UptakeData(Data):
 
         return out
 
+    @staticmethod
+    def date_to_season(date_col: pl.Expr) -> pl.Expr:
+        """
+        Extract season column from a date column, as polars expressions.
+
+        Parameters
+        date_col: pl.Expr
+            column of dates
+
+        Returns
+        pl.Expr
+            column of the season for each date
+
+        Details
+        Assume overwinter seasons, e.g. 2023-10-07 and 2024-04-18 are both in "2023/24"
+        """
+        year1 = (
+            date_col.dt.year() + pl.when(date_col.dt.month() < 7).then(-1).otherwise(0)
+        ).cast(pl.Utf8)
+        year2 = (
+            date_col.dt.year() + pl.when(date_col.dt.month() < 7).then(0).otherwise(1)
+        ).cast(pl.Utf8)
+        season = pl.concat_str([year1, year2], separator="/")
+
+        return season
+
 
 class IncidentUptakeData(UptakeData):
     def to_cumulative(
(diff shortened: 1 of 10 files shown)


iup/__init__.py
```diff
@@ -153,69 +153,63 @@ def to_incident(self, group_cols: tuple[str,] | None) -> IncidentUptakeData:
...
-            frame.select(pl.col(v) for v in group_cols.values())
-            .unique()
-            .with_columns(date=rollout, estimate=0.0)
-        )
-    else:
-        rollout_rows = pl.DataFrame({"date": rollout, "estimate": 0.0})
+        for r in rollout:
+            if group_cols is not None:
+                rollout_rows = (
+                    frame.select(pl.col(v) for v in group_cols)
```
swo, line 179:
```suggestion
                    frame.select(group_cols)
```
you can pass a list of strings to `.select()`, conveniently, so long as you're not manipulating the columns at the same time

iup/__init__.py
```diff
@@ -153,69 +153,63 @@ def to_incident(self, group_cols: tuple[str,] | None) -> IncidentUptakeData:
...
-    """
-    if group_cols is not None:
-        rollout_rows = (
-            frame.select(pl.col(v) for v in group_cols.values())
-            .unique()
-            .with_columns(date=rollout, estimate=0.0)
-        )
-    else:
-        rollout_rows = pl.DataFrame({"date": rollout, "estimate": 0.0})
+        for r in rollout:
```
swo, line 176: I think the more polaronic way to do this is:

```python
if group cols is not None:
  return frame.select(group_cols).unique().join(pl.DataFrame({'date': rollout}), how='cross').with_columns(estimate=0.0)
else:
  return pl.DataFrame({'date': rollout, 'estimate': 0.0})
```
like, there's no need to build the thing up section by section

iup/__init__.py
```diff
@@ -153,69 +153,63 @@ def to_incident(self, group_cols: tuple[str,] | None) -> IncidentUptakeData:
...
+            frame = frame.vstack(rollout_rows.select(frame.columns)).sort("date")
 
-    return frame
+        return frame
 
 
 def extract_group_names(
-    group_cols: List[dict],
-) -> tuple[str,] | None:
+    group_cols: List[List[str]],
```
swo, line 192: Is there a reason the group columns would be different by dataset?

I find this a little confusing; I would expect you to have a single dataframe, with whatever the stratifying/grouping variables are (season, geography, etc.) and then you pass that to the model

But this makes it seem like different data sets are in different data frames, with potentially different schema?

scripts/config_template.yaml
```diff
@@ -1,15 +1,19 @@
 # The data sets to load and how to interpret them
 data:
   data_set_1:
-    group_cols:
-      geography: region
-    rollout: 2023-09-01
+    rollout: [2023-09-01, 2024-09-01]
```
swo, line 4: I'm imagining some weird edge case where we inadvertently put in a rollout that is after the first data point in a season and get weird negative stuff

scripts/config_template.yaml
```diff
@@ -1,15 +1,19 @@
...
       geography_type: nation
       domain_type: age
       domain: 18+ years
       indicator_type: 4-level vaccination and intent
       indicator: received a vaccination
+      time_type: week
+    keep:
+      geography: geography
+      estimate: estimate
+      time_end: date
```
swo, line 15: I think it's actually good to keep "time_end" as the name of the column throughout the code. Because one might have a model that takes in the start and end dates? Or, at least, it removes any confusion about what the "date" is

scripts/main.py
```diff
@@ -8,21 +8,32 @@
...
+            .collect()
+            .rename(x["keep"])
+            .select(x["keep"].values())
+            .sort("date")
+        )
         for x in config["data"].values()
     ]
 
-    # List of grouping factors used in each data set
+    # Find grouping factors common to all data sets
```
swo, line 26: I'm realizing that "factor" might have confused me

E.g., we have groups of concepts like: geographic type ("state") and geography ("Wyoming"). Is Wyoming the factor, or state?

iup/__init__.py
eschrom: Also nice!

iup/__init__.py
eschrom: Different data sets are in different data frames, at least at first. E.g. if you want to train a model on covid vaccine uptake from 2022 - 2024, this spans two data sets (udsf-9v7b and ksfb-ug5d, specifically) that come out of the NIS API separately. In `main.py`, `split_train_test` is the function that concatenates them into the same frame. Before `split_train_test` is called, the data sets are in separate dataframes. So the desired grouping factors are specified for each dataframe separately (and `extract_group_names` makes sure that they match). 

Maybe a better way to do this is to have one standalone entry for desired grouping factors across all data sets in the config.yaml, rather than one under each data set? This way the user doesn't need to type out the same grouping factors repeatedly for each data set, since they must match across all data sets anyway.

scripts/config_template.yaml
eschrom: Could happen. But it is the user's responsibility to know that information. Maybe this calls for an `assert` inside `insert_rollout` that checks that the provided rollout date(s) are at least sensible.

scripts/config_template.yaml
eschrom: That's fine. Especially because `keep` doesn't have to be a dictionary, but just a list of the columns to keep, and it can be a standalone config entry instead of recurring under each data set (like the grouping factors - see above discussion). I'll just have to change some names in the data classes.

scripts/main.py
eschrom: "State" is the factor, and "Wyoming" is a level of that factor. `# Find grouping factors common to all data sets` is confusing language - what I'm trying to say is: "make sure the same columns for grouping factors are present in every data set that will contribute to the model." I will rephrase.

eschrom:
Many updates following comments from @swo, including: simpler config entries for columns to keep and grouping factor names, `date` changed to `time_end` as a column name, checking that rollout dates are at the appropriate time when uptake "starts over", etc.

iup/__init__.py
swo: `nisapi` will deliver you the stuff in udsf-9v7b and ksfb-ug5d as rows in a single dataframe, which I think means you can count on all datasets having the same schema, grouping columns, etc.

scripts/config_template.yaml
swo: or we could have `date_column` be a variable somewhere

but then maybe that needs to be at the top level of the code? which is kind of annoying. copy-paste shouldn't be too hard for now

iup/__init__.py
eschrom: I think this has been addressed in the latest updates to this PR - the user specifies which columns they want to keep (including which grouping factors they want to use), but the interface for doing so in the config is easier because we are now assuming all data sets have the same schema. `extract_group_names` is gone, actually.

iup/__init__.py
```diff
@@ -173,45 +199,37 @@ def insert_rollout(
...
+            group_cols = group_cols + ["season"]
+        else:
+            rollout_rows = pl.DataFrame({"time_end": rollout, "estimate": 0.0})
+            group_cols = ["season"]
 
-    Details
-    Before returning a single tuple of the desired column names,
-    check that they are identical for every data set.
-    """
+        frame = frame.vstack(rollout_rows.select(frame.columns)).sort("time_end")
```
swo, line 214:
```suggestion
        frame = rollout_rows.select(frame.columns).sort("time_end")
```
you don't need the vstack, since now you only have a single object!

scripts/config_template.yaml
```diff
@@ -9,11 +9,12 @@ data:
...
       indicator: received a vaccination
       time_type: week
-    keep:
-      geography: geography
-      estimate: estimate
-      time_end: date
-    group_cols: [geography]
+
+# Columns to keep from each data set
+keep: ["geography", "estimate", "time_end"]
```
swo, line 14:
```suggestion
keep: [geography, estimate, time_end]
```

swo: do we need to drop any of these other columns?

I'd suggest just letting them float along. We don't check for the *absence* of columns in any of our validations

tests/test_data_cleaning.py
```diff
@@ -28,26 +28,34 @@ def test_insert_rollout_handles_groups(frame):
...
     frame = iup.CumulativeUptakeData(frame.drop("indicator"))
 
     output = frame.insert_rollout(rollout, group_cols)
 
     assert output.shape[0] == 7
     assert (
-        output["date"].value_counts().filter(pl.col("date") == rollout[0])["count"][0]
+        output["time_end"]
+        .value_counts()
+        .filter(pl.col("time_end") == rollout[0])["count"][0]
```
swo, line 46: I think another way to do this logic is to filter and look at the number of rows:
```python
output['time_end'].filter(pl.col("time_end") == rollout[0]).shape[0] == 2
```

tests/test_linear_incident_uptake_model.py
```diff
@@ -89,7 +92,12 @@ def test_fit(frame):
     """
 
     data = iup.IncidentUptakeData(frame)
-    model = iup.models.LinearIncidentUptakeModel().fit(data, ("geography",))
+    model = iup.models.LinearIncidentUptakeModel().fit(
+        data,
+        [
+            "geography",
+        ],
```
swo, line 99:
```suggestion
        ]
```
if you cut off this comma, then ruff will probably collapse this to a single line

tests/test_linear_incident_uptake_model.py
```diff
@@ -135,7 +143,9 @@ def test_build_scaffold_handles_groups():
     start_date = dt.date(2020, 2, 1)
     end_date = dt.date(2020, 2, 29)
     interval = "7d"
-    group_cols = ("geography",)
+    group_cols = [
+        "geography",
```
swo, line 147:
```suggestion
        "geography"
```
ditto above

this is one of the weird things that I don't like, that the formatters will respect your comma as a statement that you don't want to collapse onto a single line

(I promise this is the last time I will point this out. The point of the formatter is so that we *don't* talk about style)

swo (approved):
Small ideas for style mostly

iup/__init__.py
eschrom: I think we do need the `vstack` because `rollout_rows` contains _only_ rows with 0 uptake on the rollout dates. All the other non-zero data is in `frame`. So we must stack the two together to get the full data set with rollout included.

swo: got it

tests/test_linear_incident_uptake_model.py
eschrom: No worries! I didn't know this, so it's helpful for you to point it out.

Issue #55, opened 2024-11-19
Make type hinting as precise as possible.


eschrom:
For example, Scott suggests:

`dict | None` is equivalent to `Optional[dict]`

And you could specify the key and value types with `dict[str, Any]` (or replace `Any` which whatever type you want)

eschrom:
Completed, thanks to Chris's advice.

PR #81, opened 2024-12-16
Refactor season splitter

swo:
- Pass the season start month and day as function arguments
- The unit test returns two separate values, so we make sure it isn't sending all values in the same direction

### DIFF
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -84,30 +84,38 @@ class UptakeData(Data):
         return out
 
     @staticmethod
-    def date_to_season(date_col: pl.Expr) -> pl.Expr:
+    def date_to_season(
+        date: pl.Expr, season_start_month: int = 9, season_start_day: int = 1
+    ) -> pl.Expr:
         """
-        Extract season column from a date column, as polars expressions.
+        Extract winter season from a date
+
+        Dates in year Y before the season start (e.g., Sep 1) are in the second part of
+        the season (i.e., in season Y-1/Y). Dates in year Y after the season start are in
+        season Y/Y+1. E.g., 2023-10-07 and 2024-04-18 are both in "2023/24"
 
         Parameters
-        date_col: pl.Expr
-            column of dates
+        date: pl.Expr
+            dates
+        season_start_month: int
+            month of the year the season starts
+        season_start_day: int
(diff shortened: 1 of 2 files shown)


iup/__init__.py
```diff
@@ -69,24 +69,58 @@ def split_train_test(
...
+
+        year2 = year1 + 1
+        return pl.format("{}/{}", year1, year2)
+
 
 class IncidentUptakeData(UptakeData):
     def to_cumulative(
-        self, group_cols: tuple[str,] | None, last_cumulative=None
+        self, group_cols: List[str,] | None, last_cumulative=None
     ) -> pl.DataFrame:
```
Fuhan-Yang, line 124: Why do we change from tuple to List, should `group_cols` better be immutable?

iup/__init__.py
```diff
@@ -153,71 +187,57 @@ def to_incident(self, group_cols: tuple[str,] | None) -> IncidentUptakeData:
...
-def extract_group_names(
-    group_cols: List[dict],
-) -> tuple[str,] | None:
-    """
-    Insure that the column names for grouping factors match across data sets.
+        Parameters
+        rollout: List[dt.date]
+            list of rollout dates
+        group_cols: tuple[str] | None
+            names of grouping factor columns
```
Fuhan-Yang, line 200: Then the note should be List[str] too

Fuhan-Yang (approved):
Thanks for fixing the interface with NIS data! This looks great to me.

iup/__init__.py
eschrom: The honest answer is that I'm not (yet) a stickler for the exact right data type, as long as it works. But I do take advantage of the mutability of `group_cols` in one or two places in this code.

iup/__init__.py
eschrom: Yep, I just missed it here. I'll fix up these comments when I address #55

eschrom (approved):
I like the more flexible version of `date_to_season`! Looks good to me.

iup/__init__.py
swo: These changes were part of a previous PR; I just rebased; so you can see @Fuhan-Yang that this PR is really just the 1 commit

PR #82, opened 2024-12-16
Add evaluation to configuration

Fuhan-Yang:
This PR adds time-varying evaluation, updates README, and refactors test file. 
Evaluation is conducted in the same `timeframe` as projections. It supposes to evaluate the metrics sequentially as forecast start date moving forward by a given interval, ending one interval before forecast end date. `option` is added to enable the choice between producing projections or evaluation. 

More details are added to Getting Started in README. 

Evaluation test is simplified to one test function.


tests/test_evaluation.py
```diff
@@ -39,39 +39,24 @@ def pred():
...
+def test_score_df(data, pred, score_fun):
     """
-    Return the expected forecast start, end and correct MSPE.
+    Return the expected forecast start, end, and correct MSPE.
     """
     data = iup.IncidentUptakeData(data)
     pred = iup.PointForecast(pred)
 
-    output = eval.score(data=data, pred=pred, score_fun=eval.mspe)
+    output = eval.score(data=data, pred=pred, score_fun=score_fun)
```
eschrom, line 49: How does the argument `score_fun` get its value? I don't see it assigned anywhere, nor do I see a default value assigned in `def score` in `iup/eval.py`. Maybe you need an `@pytest.fixture` for `score_fun` to pass it into your test this way?

tests/test_evaluation.py
```diff
@@ -39,39 +39,24 @@ def pred():
...
-    Return the expected forecast start, end and correct mean bias.
-    """
-    x = pl.Series([0.0, 0.1, 0.7, 0.4, 0.5])
-    y = pl.Series([0.0, 0.2, 1.0, 0.6, 0.5])
-    assert np.isclose(eval.mean_bias(x, y), -0.6)
+    # Validate score type
+    assert isinstance(output["score"][0], float)
 
+    # Expected values for different scoring functions
+    expected_scores = {eval.mspe: 0.028, eval.mean_bias: -0.6, eval.eos_abe: 0.352941}
```
eschrom, line 59: Looks like multiple small tests have been condensed into one larger test. I was under the impression that multiple tests are actually better, because they help you pinpoint exactly what is failing when something fails. For example, if this test fails, you might not know if the problem is in the `score()` wrapper vs. the more granular score-calculator functions. In the latter case, depending on how `score_fun` is passed as an argument, you may not even know whether the `mspe` vs. `mean_bias` vs. `eos_abe` is the problem. But maybe @swo can comment more on the optimal design of these tests.

eschrom (changes requested):
I have marked a couple spots that may need to be addressed: refitting the model after each increment of the projection start date, and some pytest design stuff. 

But overall, this is a really nice integration of the model building vs. evaluation code, in both the config and in main.py. This helps me see how the two halves of the code fit together. Thanks, Fuhan!

scripts/main.py
```diff
@@ -37,15 +41,74 @@ def run(config: dict, cache: str):
...
-        config["timeframe"]["start"],
-        config["timeframe"]["end"],
-        config["timeframe"]["interval"],
-        grouping_factors,
-    )
-    print(cumulative_projections)
-    incident_projections = cumulative_projections.to_incident(grouping_factors)
-    print(incident_projections)
+
+    option = config["option"]
```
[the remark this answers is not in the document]
Fuhan-Yang: This sounds good too!

scripts/main.py
```diff
@@ -37,15 +41,74 @@ def run(config: dict, cache: str):
...
+            cumulative_projections = incident_model.predict(
+                date,
+                config["timeframe"]["end"],
+                config["timeframe"]["interval"],
+                grouping_factors,
+            )
+
+            pred = iup.PointForecast(
+                cumulative_projections.to_incident(grouping_factors)
+            )
```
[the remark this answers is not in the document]
Fuhan-Yang: This is a good point. Here I evaluated retrospective forecasting, which is evaluating the model performance as the forecast period changes. This requires the model stays the same to objectively observe if the model forecasts better when changing the forecast time length, lead time (forecast start). Thus the training data needs to be the same to have the same model. While adding data to training sequentially as forecasting moving forward is also valid, which is called real-time forecasting(I think?). This is more commonly applied in operational forecasting in practice. Which method to use depends on the scope of this project, I'd like to hear about what @swo thinks!

tests/test_evaluation.py
Fuhan-Yang: This is a good catch! The bug has been fixed.

tests/test_evaluation.py
Fuhan-Yang: This is a good point. I combined these three tests followed the [comment]( https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/73#discussion_r1874057551). I agree that it is more clear to have granular tests.

scripts/main.py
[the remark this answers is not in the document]
swo: To make sure we're on the same page, some definitions:

- Forecast date: The date the forecast was run. In a retrospective analysis, you run a forecast with the data that would have been available on this date.
- Observation date: The date that the thing you're trying to forecast occurred.
- Forecast horizon: Time between the forecast date and the observation date. E.g., if a Jan 1 forecast is predicting what will happen on Feb 1, that's a 1 month horizon.

For a given forecast date, estimates on observation dates that are further into the future (e.g., at longer forecast horizons) will have lower accuracy.

So I think @eschrom is saying that, in a retrospective analysis, you run forecasts on many forecast dates. Later forecast dates have access to more data. The data that comes out of a retrospective analysis will be many combinations of forecast dates and observation dates. Then you would score a model by saying, e.g., what was its accuracy at a 1-month forecast horizon, averaged over all the relevant forecast dates.

image

Does this help at all?

README.md
```diff
@@ -16,11 +16,16 @@ Use <https://github.com/CDCgov/nis-py-api> for access to the NIS data.
...
-5. `make run`
+1. Set up a virtual environment with `poetry shell`
+2. Installed the required dependencies with `poetry install`
+3. Get a [Socrata app token](https://github.com/CDCgov/nis-py-api?tab=readme-ov-file#getting-started) and save it in `scripts/socrata_app_token.txt`
+4. Cache NIS data with `make cache`
+5. Copy the config template in `scripts/config_template.yaml` (e.g., to `scripts/config.yaml`) and fill in the necessary fields
+    - data: specify the type of vaccine data in terms of: geography, demography (domain), and vaccine status(indicator).
+    - timeframe: specify the start and the end of time in predicted projections or evaluation.
+    - option: specify the returned output. Can be either "prediction" or "evaluation".
+    - metrics: specify the evaluation metrics. Only valid when option is "evaluation". Can be one of "mspe", "mean_bias","end_of_season_error", or "all".
```
swo, line 27: This would be better as a list? Right now your options are 1 or a particular string, but better to just let the user pick what they want

scripts/config_template.yaml
```diff
@@ -21,3 +21,10 @@ timeframe:
   start: 2024-02-01
   end: 2024-04-30
   interval: 7d
+
+
+# The options to return projection or evaluation metrics, can be "projection" or "evaluation"
+option: projection
```
swo, line 27: I think you want both? Like, I'm never going to want *just* scores, I'm always going to want to see scores *and* some diagnostic plots

Also, don't you need the projections, to do the scores? So one of these options is "do the projections, get the scores, and throw the projections away"

I think this is where you'd be better off having a workflow that makes the projections, and then builds off that to do the evaluation. You can code up that workflow in a Makefile

scripts/main.py
```diff
@@ -46,15 +49,74 @@ def run(config: dict, cache: str):
...
+            test_data = iup.IncidentUptakeData(
+                iup.IncidentUptakeData.split_train_test(
+                    incident_data, date, "test"
+                ).filter(pl.col("date") <= config["timeframe"]["end"])
+            )
+
+            if config["metric"] == "mspe":
```
swo, line 105: Rather than these if-else's, you could do something like:

```python
score_fun = getattr(eval, config['metric'])
score = eval.score(test_data, pred, score_fun)
```

Essentially you grab the function based on its string name

scripts/main.py
```diff
@@ -46,15 +49,74 @@ def run(config: dict, cache: str):
...
+        scores = pl.DataFrame(
+            schema={
+                "forecast_start": pl.Date,
+                "forecast_end": pl.Date,
+                "score": pl.Float64,
+                "type": pl.String,
+            }
+        )
+
+        for date in dates:
```
swo, line 87: I strongly recommend you move away from for loops and toward functions here

This is where the Makefile or Snakemake will come in handy. Like, you're doing many different jobs:

- For each combination of forecast date and model, do the forecast
- For each model and score function (and forecast horizon?), score the model using the forecasts from the previous step
- Report these out in a table?
- Make some figures?

So rather than saying "what do I need to do for each date?" and putting that into a for loop, instead say "what are the different jobs I need to do?" and write a function for each one.

There will be looping logic somewhere, but it might be in the Makefile or snakefile

tests/test_evaluation.py
swo: I think you want to keep the tests as they currently are on main, that just take in an `x` and `y` vector and return a value. These are very easy to interpret

You might also want to check that `eval.score` can take a score function, etc., but you can decouple that logic from whether `eos_abe` gives you the correct numerical result

swo (changes requested):

scripts/config_template.yaml
eschrom: I agree with @swo. Instead of having options at all, I recommend just one workflow that creates projections and gives the desired scores. 

Specifically, your `interval_number` in main.py line 69 could be set by the user in the config, rather than taken from `config["timeframe"]["interval"]`. This would allow the user to customize the forecasts independently of the time intervals in the data. That is, data points are reported every week, but your forecast dates could be, e.g., monthly with forecasts starting on Feb 1, Mar 1, Apr 1, etc. If the user doesn't enter a value for this in the config, the default behavior could be to just generate one forecast and evaluation, without varying the forecast date at all.

I like having one script, `main.py`, that runs the entire pipeline every time, because this helps me check that all parts of the pipeline cooperate with each other. But ultimately the Makefile will be better because it won't repeat stuff that's already been done and is up-to-date. Perhaps we should work together on such a Makefile in a separate PR.

scripts/main.py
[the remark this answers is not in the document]
eschrom: I guess this is superseded by the fact that I agree with @swo above.

scripts/main.py
eschrom: Why is it better to have looping logic in a Makefile than in the script? Just so that iterations of the loop that have already been completed aren't repeated?

scripts/main.py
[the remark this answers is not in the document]
eschrom: I had not been familiar with the vocabulary, so this helps a lot. I like the Paint-style sketch, too ;)

"In a retrospective analysis, you run a forecast with the data that would have been available on this date ...  in a retrospective analysis, you run forecasts on many forecast dates. Later forecast dates have access to more data" This is exactly what I am saying, but more eloquent.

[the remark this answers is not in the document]
Fuhan-Yang: But how do we objectively evaluate the performance if continuously adding data to training? In this way, the forecasts made on later forecast dates must be more accurate than the ones made on early forecast dates only because they have more training data, and we cannot evaluate the relation between forecast performance and forecast dates (such as if approaching a seasonal onset affect performance) unbiasedly.

tests/test_evaluation.py
Fuhan-Yang: Just checked `test_score_df` independently and it worked, this also passed pytest so maybe `eval.score` can take a score function. Does that mean we still keep the granular tests but just replace `x` and `y` with `data` and `pred`?

scripts/main.py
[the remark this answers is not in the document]
eschrom: You are correct that "forecasts made on later forecast dates must be more accurate than the ones made on early forecast dates only because they have more training data." But this is not a problem. 

We want to compare forecasts between _models_, not between _forecast dates_. We don't care that a given model always performs better for later forecast dates. We care whether one model performs better than another at a given forecast date.

Here's a sketch of my own:
image

Both models improve in performance as forecast date advances. But the key is to compare the models. In this case, we see that Model 2 performs better for early forecast dates, but Model 1 performs better for later forecast dates.

swo:
@eschrom @Fuhan-Yang My top line point in all of this is that we would need to loop over a lot of things:

- Data set
- Season (although I'm confused if "data set" and "a season of data for a particular vaccine" are different things)
- Model
- Forecast date
- Score function (aka evaluation metric)

See one example of what that would look like, using loops, in #83 

I think the way you want to do this instead is:

1. Preprocessing
   - Clean the raw data
   - Separate the raw data into whatever chunks you'll want to run independently (i.e., a season of data for a particular vaccine) -- this could still be one data frame, but with some column that allows you to pick out the particular chunk you want to use
   - Save this somewhere
2. Model runs
   - For every "chunk" and forecasting date and model, do the forecasts
   - I think it's way easier to think of this, having a `forecast()` function somewhere in there, rather than computation nested inside 3 levels of looping!
   - Save every result
3. Evaluation
   - For every chunk, model, forecast date, and score function, do the scores
   - Save every result

Unfortunately Makefiles don't make it easy to say "run every model on every forecast date." (Something like Snakemake could, but let's leave that aside for now.) So I don't mind sticking with Python for now.

Unless this whole pipeline runs in less than 5 seconds, I would want preprocessing, doing the forecasts, and scoring the forecasts to be in separate scripts. Otherwise the time required to debug and iterate is going to be too long!

swo:
And then see #84 for how I would (start to) adapt those loops into those separate scripts

scripts/main.py
Fuhan-Yang: Does for loop play as a high-level command so it should be in the Makefile, while low-level command should be functions simple enough to debug?

scripts/config_template.yaml
Fuhan-Yang: I will leave this in a future PR.

scripts/main.py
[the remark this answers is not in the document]
Fuhan-Yang: See [b127d43](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/82/commits/b127d43fe50e80b1f31a684d7a3c6e0596265c17)

scripts/main.py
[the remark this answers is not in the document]
Fuhan-Yang: See [b127d43](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/82/commits/b127d43fe50e80b1f31a684d7a3c6e0596265c17)

scripts/config_template.yaml
Fuhan-Yang: Will leave this to the next PR.

scripts/main.py
Fuhan-Yang: See [b127d43](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/82/commits/b127d43fe50e80b1f31a684d7a3c6e0596265c17)

tests/test_evaluation.py
Fuhan-Yang: See [dad1552](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/82/commits/dad15520418db893fa05b1faec026f4aa126865e)

README.md
Fuhan-Yang: New issue created. https://github.com/CDCgov/cfa-immunization-uptake-projection/issues/85#issue-2748543790

scripts/main.py
Fuhan-Yang: New issue created https://github.com/CDCgov/cfa-immunization-uptake-projection/issues/86#issue-2748546812

scripts/main.py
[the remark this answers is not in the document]
eschrom: This adjustment looks right! I think this is the crucial thing to fix in this PR.

scripts/main.py
eschrom: I agree it's fine to postpone the relocation of looping for the next PR, because it could be influenced by our redistribution of `main.py` into 3-4 separate scripts.

eschrom (approved):
Looks like the major discussion point for this PR - retraining the model as the forecast date moves - is fixed!

scripts/config_template.yaml
swo: Punting to #89

swo (approved):
We're moving in the right direction! Let's get this merged and move on to the next steps

PR #83, opened 2024-12-18
Pipeline setup

swo:
Note: The scripts here are not functional. There's even malformed Python code; this PR is intended to be fixed by later ones right away.

- Read model names and score function names from config
- Do computation for every combination of data set, model, forecast date, and score function
- Separate into scripts
- Use a Makefile


Fuhan-Yang:
This PR seperates one main.py to three files as preprocess.py, forecast.py, and eval.py. Each file receives the input from config.yaml, and is called by Makefile sequentially. To test the new pipeline, several errors need to fix:

1. Error: "typeFloat64 is incompatible with expected type String" is returned in `split_train_test`. This happens when the uptake data is a single polars DataFrame (instead of a uptake data list). 
2. The NaN in training data prevents the fitting of LinearIncidenceUptakeModel, checks or methods (impute NA, or delete) are needed to ensure fitting can be run.

Fuhan-Yang (approved):

PR #92, opened 2025-01-10
Pipeline setup step 1: Fix preprocess to run Makefile

Fuhan-Yang:
This PR fixes some errors when calling `make all`. The changes are only made to run the code successfully, and not guarantee if it introduces errors in the output.

### DIFF
--- a/scripts/preprocess.py
+++ b/scripts/preprocess.py
@@ -29,15 +29,7 @@ def preprocess(
         cumulative_data.insert_rollout(rollout_dates, groups)
     )
 
-    # Convert to incident data
-    incident_data = cumulative_data.to_incident(groups)
-
-    return pl.concat(
-        [
-            cumulative_data.with_columns(estimate_type=pl.lit("cumulative")),
-            incident_data.with_columns(estimate_type=pl.lit("incident")),
-        ]
-    )
+    return cumulative_data
 
 
 if __name__ == "__main__":
--- a/scripts/preprocess.py
+++ b/scripts/preprocess.py
@@ -18,7 +18,7 @@ def preprocess(
 ) -> pl.DataFrame:
     # Prune data to correct rows and columns
     cumulative_data = iup.CumulativeUptakeData(
-        raw_data.filter(filters).select(keep).sort("time_end").collect()
+        raw_data.filter(**filters).select(keep).sort("time_end").collect()
     )
 
     # Ensure that the desired grouping factors are found in all data sets
@@ -34,8 +34,8 @@ def preprocess(
 
     return pl.concat(
         [
-            cumulative_data.with_columns(estimate_type="cumulative"),
-            incident_data.with_columns(estimate_type="incident"),
+            cumulative_data.with_columns(estimate_type=pl.lit("cumulative")),
+            incident_data.with_columns(estimate_type=pl.lit("incident")),
         ]
     )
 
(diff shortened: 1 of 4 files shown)


scripts/preprocess.py
```diff
@@ -59,10 +61,10 @@ def preprocess(
 
     clean_data = preprocess(
         raw_data,
-        filters=config["data"][0]["filters"],
-        keep=config["data"][0]["keep"],
+        filters=config["data"]["data_set_1"]["filters"],
```
eschrom, line 64: This script assumes there will be only one data set described in the config (see "data_set_1" here and also "Don't know how to preprocess multiple data sets" in line 58). But this script _must_ be able to preprocess multiple data sets. Models should be fit on _all_ previous data for the immunization of interest (covid booster or flu vaccine), and that previous data is split into two separate data sets by the NIS API (e.g. ksfb-ug5d is covid booster for 2023-24, whereas udsf-9v7b is covid booster for pre-2023).

scripts/preprocess.py
```diff
@@ -34,8 +34,8 @@ def preprocess(
 
     return pl.concat(
         [
-            cumulative_data.with_columns(estimate_type="cumulative"),
-            incident_data.with_columns(estimate_type="incident"),
+            cumulative_data.with_columns(estimate_type=pl.lit("cumulative")),
+            incident_data.with_columns(estimate_type=pl.lit("incident")),
```
eschrom, line 38: Previously, we've been very careful to separate `CumulativeUptakeData` from `IncidentUptakeData`, so why are we combining both into a single data frame here? Why not just return the cumulative uptake data, since this is easily converted to incident during any future step?

eschrom (changes requested):
Thanks, @Fuhan-Yang, for tackling the reorganization into separate scripts! I missed out on reviewing #83 so some of my comments go back to changes made there. Broadly, I think we should be careful that whatever we merge not only runs without runtime errors, but also does not introduce logical problems, passes pre-commit, etc. Currently, this PR inherits from #83 some logical oddities, which I have called attention to, and some pre-commit issues.

swo:
#95 should fix pre-commit compliance here

scripts/preprocess.py
swo: Agree that cumulative uptake data is what should come from upstream, go through pre-processing, and be passed to the models.

Some models don't need incident data. Those that do should be forced to choose things like the rollout date and how to convert cumulative CIs into incident CIs, which could be considered a non-trivial choice (and therefore at the level of a model, rather than pre-processing).

scripts/preprocess.py
swo: I don't know that we need to divide into "datasets" any more. `nisapi` will give you one giant data frame with all pathogens and seasons.

eschrom: How do you get one giant data frame? Last I used the NIS API, I got 4 separate .parquet files: recent covid, historical covid, recent flu, historical flu.

scripts/preprocess.py
Fuhan-Yang: Fixed in [8cf3aea](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/92/commits/8cf3aeaef92183f510f6ebc781de28a665642767)

scripts/preprocess.py
eschrom: I think I see now - NIS API does create four .parquet files separated by pathogen and season, but then using `get_nis` on the path that includes all four files will load them together as a single giant data frame combining all pathogens and seasons. Is that correct, @swo?

eschrom (approved):
I see where placeholders have been added to appease pre-commit, and that only cumulative data is passed along. I also think I now understand how the NIS API returns one giant data set - this will necessitate #96

scripts/preprocess.py
swo: Correct, the files are partitions of the data set. Because they all have the same schema, and the folders are named with Hive-style naming, we can grab them all at once: https://docs.pola.rs/user-guide/io/hive/

swo (approved):
We're moving in the right direction, but I think we need #98 to resolve some of the things we discussed synchronously

PR #93, opened 2025-01-10
Pipeline setup step 2: Fill forecast.py

Fuhan-Yang:
This PR adds forecast function and run the forecast sequentially given the time frame in config.yaml. The results are saved as parquet data in the designated path. While this function hasn't been tested because of several errors: 

1. Error: "typeFloat64 is incompatible with expected type String" is returned in `split_train_test`. This happens when the uptake data is a single polars DataFrame (instead of a uptake data list).

2. The NaN in training data prevents the fitting of `LinearIncidenceUptakeModel`, checks or methods (impute NA, or delete) are needed to ensure fitting can be run.

### DIFF
--- a/scripts/forecast.py
+++ b/scripts/forecast.py
@@ -12,6 +12,9 @@ def run_all_forecasts(clean_data, config) -> pl.DataFrame:
     Returns:
         pl.DataFrame: data frame of forecasts, organized by model and forecast date
     """
+    raise NotImplementedError
+    models = None
+
     forecast_dates = pl.date_range(
         config["timeframe"]["start"],
         config["timeframe"]["end"],
@@ -64,7 +67,7 @@ def run_forecast(
     )
 
     # Fit models using the training data and make projections
-    fit_model = model().fit(incident_train_data, grouping_factors)
+    # fit_model = model().fit(incident_train_data, grouping_factors)
 
     cumulative_projections = fit_model.predict(
         forecast_start,
--- a/scripts/forecast.py
+++ b/scripts/forecast.py
@@ -73,10 +73,7 @@ def run_forecast(
         grouping_factors,
     )
 
-    incident_projections = cumulative_projections.to_incident(grouping_factors)
-
-    # Note that here returns incident projections only, for evaluation
-    return incident_projections
+    return cumulative_projections
 
 
 if __name__ == "__main__":
--- a/scripts/forecast.py
+++ b/scripts/forecast.py
@@ -12,7 +12,7 @@ def run_all_forecasts(clean_data, config) -> pl.DataFrame:
     Returns:
         pl.DataFrame: data frame of forecasts, organized by model and forecast date
     """
-    
+
     forecast_dates = pl.date_range(
         config["timeframe"]["start"],
         config["timeframe"]["end"],
@@ -59,13 +59,13 @@ def run_forecast(
     incident_data = iup.CumulativeUptakeData(observed_data).to_incident(
         grouping_factors
     )
-    
+
     incident_train_data = iup.IncidentUptakeData(
         iup.IncidentUptakeData.split_train_test(incident_data, forecast_start, "train")
     )
 
     # Fit models using the training data and make projections
-    # fit_model = model().fit(incident_train_data, grouping_factors)
+    fit_model = model().fit(incident_train_data, grouping_factors)
 
     cumulative_projections = fit_model.predict(
         forecast_start,
--- a/scripts/forecast.py
+++ b/scripts/forecast.py
@@ -3,7 +3,10 @@ import argparse
 import polars as pl
 import yaml
 
-def run_all_forecasts() -> pl.DataFrame:
+import iup.models
+
+
+def run_all_forecasts(clean_data, config) -> pl.DataFrame:
     """Run all forecasts
 
     Returns:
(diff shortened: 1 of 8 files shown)


eschrom (changes requested):
Similar to my review for #92, I think we should get this to pass pre-commit and root out errors before merging.

If `split_train_test` breaks when given a single polars data frame, let's fix it. It looks like preprocess.py intends to always return a single data frame, not a list. This can easily remain true even after preprocess.py knows how to process multiple data sets, because it can just concatenate them into one before returning it. So we can just edit `split_train_test` to expect only a single data frame.

Likewise, if NaNs in the data cause problems, let's find out where they come from and fix them if possible. I don't recall missing values in the raw data - I wonder if there are further issues lurking in our code that cause the NaNs.

Normally, such things could be separate issues for separate PRs, but if they prevent us from testing the content of this PR properly, then I think they should be solved here.

Please also see my separate comment on the way to increment the forecast date.

scripts/forecast.py
```diff
@@ -35,36 +64,33 @@ def run_forecast() -> pl.DataFrame:
...
-        iup.IncidentUptakeData.split_train_test(
-            incident_data, config["timeframe"]["start"], "test"
-        )
-    ).filter(pl.col("date") <= config["timeframe"]["end"])
+    return pl.concat(
+        [
+            cumulative_projections.with_columns(estimate_type=pl.lit("cumulative")),
+            incident_projections.with_columns(estimate_type=pl.lit("incident")),
+        ]
+    )
```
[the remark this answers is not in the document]
Fuhan-Yang: Fixed in [862faf8](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/93/commits/862faf83e7afa9a4a04282fb32bfccb58007a242)

scripts/forecast.py
```diff
@@ -18,13 +21,39 @@ def run_all_forecasts() -> pl.DataFrame:
...
-def run_forecast() -> pl.DataFrame:
+    for model in models:
+        for forecast_date in forecast_dates:
+            # Get data available as of the forecast date
+            forecast = run_forecast(
+                model,
+                clean_data,
+                grouping_factors=config["groups"],
+                forecast_start=config["timeframe"]["start"],
+                forecast_end=forecast_date,
```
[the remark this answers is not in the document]
Fuhan-Yang: Fixed in [857c080](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/93/commits/857c0808b8a47e2dfc6e2109dc5469ac50c5f6c2)

Fuhan-Yang:
> Similar to my review for #92, I think we should get this to pass pre-commit and root out errors before merging.
> 
> If `split_train_test` breaks when given a single polars data frame, let's fix it. It looks like preprocess.py intends to always return a single data frame, not a list. This can easily remain true even after preprocess.py knows how to process multiple data sets, because it can just concatenate them into one before returning it. So we can just edit `split_train_test` to expect only a single data frame.
> 
> Likewise, if NaNs in the data cause problems, let's find out where they come from and fix them if possible. I don't recall missing values in the raw data - I wonder if there are further issues lurking in our code that cause the NaNs.
> 
> Normally, such things could be separate issues for separate PRs, but if they prevent us from testing the content of this PR properly, then I think they should be solved here.
> 
> Please also see my separate comment on the way to increment the forecast date.

For now, I would suggest to merge the PRs when the questions are well-addressed. Once the entire pipeline is merged, we can test it and open PRs if needed!

scripts/forecast.py
```diff
@@ -72,12 +75,8 @@ def run_forecast(
...
     incident_projections = cumulative_projections.to_incident(grouping_factors)
 
-    return pl.concat(
-        [
-            cumulative_projections.with_columns(estimate_type=pl.lit("cumulative")),
-            incident_projections.with_columns(estimate_type=pl.lit("incident")),
-        ]
-    )
+    # Note that here returns incident projections only, for evaluation
+    return incident_projections
```
eschrom, line 79: I think we should return cumulative projections here. The reason is that incident data for a partial season loses information. Suppose you are making forecasts from Jan 1 - May 31. Cumulative forecasts contain all the incident information, but incident forecasts "forget" what the total cumulative uptake had been up to Jan 1. So if you convert to incident here, it will be harder to convert back to cumulative later (e.g. for calculating end-of-season absolute error) - you'll have to look up the observed data again. But if you keep the cumulative version here, it is easy to convert to incident later (e.g. for calculating mspe).

eschrom (changes requested):
I added one more comment regarding cumulative vs. incident. I think it will be useful to only pass cumulative data around, converting to incident only at the point where incident is needed (e.g. fitting the linear model, calculating mspe, etc.). But take a look and let me know what you think!

scripts/forecast.py
swo: I agree. After all our discussions, my thinking now is that the inputs and outputs from the models should be in cumulative space (although many of the models might internally convert back and forth with incident)

eschrom: I just looked back at your eval.py code, and I realized that end-of-season absolute error is, more specifically the error in cumulative uptake _since the forecast date_. There is no error before the forecast date - that data is known. So your approach makes more sense to me now.

I still propose we follow the convention only to pass cumulative data around and only to convert to incident at the points it's required, but I wanted you to know that I resolved some of my own confusion!

swo (commented):
I'd like to resolve #92 and #98 before working deeply on this one

Also, it would be nice if this were a PR into #92's branch, rather than into `main`, so that we see only the incremental changes that this makes (and so we don't see, e.g., the changes in `eval.py`)

scripts/forecast.py
Fuhan-Yang: Agree that cumulative projections are more informative and should pass along. Just fixed this [c2cbdd0](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/93/commits/c2cbdd029102aba9e08a07743ecea7532e609eb8). As #92 and #98 are merged, I think we can merge this one if there is no other issue.

Fuhan-Yang:
> I'd like to resolve #92 and #98 before working deeply on this one
> 
> Also, it would be nice if this were a PR into #92's branch, rather than into `main`, so that we see only the incremental changes that this makes (and so we don't see, e.g., the changes in `eval.py`)

Sure, I will use this recursive branching method when it comes to incremental change in the future!

eschrom (approved):
Looks good to me.

PR #94, opened 2025-01-10
Pipeline setup step 3: Add eval.py

Fuhan-Yang:
This PR add the last functionality in the pipeline, evaluation. The metric will be calculated for different forecast end dates and for each model. This PR hasn't been tested with real data yet, i.e, the pipeline hasn't been run entirely.

### DIFF
--- a/scripts/eval.py
+++ b/scripts/eval.py
@@ -7,16 +7,14 @@ import iup
 from iup import eval
 
 
-def eval_all_forecasts(test, pred, config):
+def eval_all_forecasts(data, pred, config):
     """Evaluate the forecasts for all models, all forecast ends, and all scores"""
     score_names = config["score_funs"]
     model_names = pred["model"].unique()
     forecast_starts = pred["forecast_start"].unique()
 
     # only 'incident' type is evaluated #
-    incident_pred = pred.filter(pl.col("estimate_type") == "incident").with_columns(
-        quantile=0.5
-    )
+    incident_pred = pred.with_columns(quantile=0.5)
     # This step is arbitrary, but it is necessary to pass PointForecast validation #
 
     all_scores = pl.DataFrame()
@@ -28,6 +26,11 @@ def eval_all_forecasts(test, pred, config):
             for forecast_start in forecast_starts:
                 pred_data = incident_pred.filter(
                     pl.col("model") == model, pl.col("forecast_start") == forecast_start
+                )
+
+                test = data.filter(
+                    pl.col("time_end") >= forecast_start,
+                    pl.col("time_end") < config["timeframe"]["end"],
                 )
 
                 assert (pred_data["forecast_start"] == test["time_end"].min()).all()
@@ -56,14 +59,9 @@ if __name__ == "__main__":
 
     pred_data = pl.scan_parquet(args.pred).collect()
     obs_data = pl.scan_parquet(args.obs).collect()
-    obs_data = obs_data.filter(pl.col("estimate_type") == "incident")
 
     # ensure the same test data is used for all models
-    test_data = iup.IncidentUptakeData.split_train_test(
-        obs_data, config["timeframe"]["start"], "test"
-    ).filter(pl.col("time_end") <= config["timeframe"]["end"])
+    obs_data = iup.CumulativeUptakeData.to_incident(config["groups"])
 
-    test_data = iup.IncidentUptakeData(test_data)
-
-    all_scores = eval_all_forecasts(test_data, pred_data, config)
+    all_scores = eval_all_forecasts(obs_data, pred_data, config)
     all_scores.write_parquet(args.output)
--- a/scripts/eval.py
+++ b/scripts/eval.py
@@ -13,10 +13,6 @@ def eval_all_forecasts(data, pred, config):
     model_names = pred["model"].unique()
     forecast_starts = pred["forecast_start"].unique()
 
-    # only 'incident' type is evaluated #
-    incident_pred = pred.with_columns(quantile=0.5)
-    # This step is arbitrary, but it is necessary to pass PointForecast validation #
-
     all_scores = pl.DataFrame()
 
     for score_name in score_names:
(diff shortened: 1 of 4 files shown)


scripts/eval.py
```diff
@@ -1,6 +1,69 @@
...
+
+    with open(args.config, "r") as f:
+        config = yaml.safe_load(f)
+
+    pred_data = pl.scan_parquet(args.pred).collect()
+    obs_data = pl.scan_parquet(args.obs).collect()
+    obs_data = obs_data.filter(pl.col("estimate_type") == "incident")
+
+    # ensure the same test data is used for all models
+    test_data = iup.IncidentUptakeData.split_train_test(
```
eschrom, line 62: I think there will need to be a loop outside of `eval_all_forecasts` to split out the test data differently for each different `forecast_start` date, once #93 iterates over `forecast_start` instead of `forecast_end`.

eschrom (changes requested):
My comments here are largely just the fallout of comments on #92 and #93. Plus, I think it's also worth debating whether we should be evaluating incident vs. cumulative uptake.

scripts/eval.py
```diff
@@ -1,6 +1,69 @@
...
+from iup import eval
+
+
+def eval_all_forecasts(test, pred, config):
+    """Evaluate the forecasts for all models, all forecast ends, and all scores"""
+    score_names = config["score_funs"]
+    model_names = pred["model"].unique()
+    forecast_ends = pred["forecast_end"].unique()
+
+    # only 'incident' type is evaluated #
```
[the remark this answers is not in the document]
Fuhan-Yang: I think it depends on the metrics we use. For example, MSPE will strictly requires the expected values of the model output for calculation (see definition [here](https://en.wikipedia.org/wiki/Mean_squared_prediction_error)). This means we should only use the direct output (incident data) to get MSPE, without modifying anything. Otherwise, we are not correct in using MSPE. However, we can make our interested metrics and use cumulative data. For example, we also calculated the absolute error of end-of-season uptake. I used the cumulative projections (from the incident projections) and the cumulative test data to calculate that. For the code structure, here we used incident projections as input for all the metrics, and we calculate the cumulative projections within the metric function if it's needed. I would suggest we keep the existing structure, and we can always add the interested metrics about cumulative projections as we go on!

scripts/eval.py
eschrom: But `eval_all_forecasts` accepts exactly one test data set, right? As `forecast_start` changes, the test data set also changes (for later `forecast_start` dates, the test data set shrinks, just like the training data set grows). So as we iterate over different `forecast_start` dates, we'll need different test data sets, which must be given to `eval_all_forecasts` individually. That's why it seems to me like the loop over `forecast_start` must be outside `eval_all_forecasts`. Does that make sense?

Fuhan-Yang: That's a good catch! Yes, I should separate the test data as the same length as the predictions. How about inputting the entire data set, and filtering the test data given the same time period as each prediction, in the most inner layer of the for loop (between line 29 and 41)?

eschrom: Sure, that sounds like an equally fine solution!

Fuhan-Yang: See [83d5341](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/94/commits/83d53413aab36b785544033b34fdc191cb44b09a)

swo:
> I think it's also worth debating whether we should be evaluating incident vs. cumulative uptake.

Some metrics will clearly be based on cumulative uptake (e.g., end-of-season uptake). For scores like mspe, it seems like incident space is better (so that you're not penalizing the model every week for a single mistake in incident uptake that's led to an ongoing bias in cumulative uptake).

Although the final answer might be that we want both: if there's a model that performs well in incident space but not cumulative space (or vice versa), that's interesting & important. But if the models have the same *relative* performance, whether using incident or cumulative, then it's an academic question

scripts/eval.py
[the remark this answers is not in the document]
swo: Some metrics will clearly be based on cumulative uptake (e.g., end-of-season uptake). For everything else, I think we'll want to score both incident and cumulative. We could consider those as pairs of metrics, e.g., we have incident MSPE and cumulative MSPE.

If there's a model that performs well in incident space but not cumulative space (or vice versa), that's interesting & important and we would have missed it by picking to evaluate one or the other.

If the models have the same *relative* performance, whether using incident or cumulative, then it's an academic question and we can pick either one.

scripts/eval.py
swo: I think there might be some confusion here about what "forecast start" and "forecast end" mean?

My understanding was that the "timeframe" parameter in the config was a series of *forecast dates*. On each forecast date, we make a forecast for every *target date*. The target dates are every week from the forecast date to the end of the season.

eschrom: You are correct about the "timeframe" config parameters: "start" is the forecast date, "end" is the end-of-season, and target dates are (currently set to) weekly in between. So forecasts are trajectories from "start" to "end." But @Fuhan-Yang is also interested in how model performance changes as "start" is pushed back later in the season. In the config, you can make "timeframe" "start" a list of dates, rather than a single date, to generate multiple forecast trajectories that start on different dates.

swo: Yes, in general we should expect better performance for shorter *forecast horizons* (i.e., time from forecast date to target date). The later in the season the forecast date, the better the prediction of the end-of-season uptake.

Fuhan-Yang: "forecast start" here is the next day of the end of training period and the day of the beginning of forecast. As "forecast start" incrementally approaches to "forecast end" (which is the end of season by default) done by `run_all_forecast`(), we can see how the performance changes along time.

Fuhan-Yang:
Just updated the eval.py to be compatible with the returned cumulative predictions. See [f9f1526](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/94/commits/f9f1526f73328afdd3cc5181de3df4966b377f00)

eschrom (approved):
I think all the changes in the eval script make sense to me now, although I must admit it is hard to imagine possible bugs without running the whole pipeline. I think we should merge #93 and #94, and then I volunteer to debug the pipeline from start to finish.

scripts/eval.py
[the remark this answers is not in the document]
Fuhan-Yang: Just opened issue #100 to record the evaluation metrics. I will add the metrics of sample distribution in the future. Feel free to add the metric-of-interest. While for MSPE, I still think it's not feasible for cumulative predictions. It is because we optimized the model using ordinary least-square method, reducing the difference between the fitted line and the training data. So it is natural to use the same measure on the prediction and the testing data. I think MSPE should be consistent with the training data and the prediction. That said, we shouldn't use incident data to train the model but evaluate it with cumulative data. The mismatch of the data types may lead to ambiguous results and challenges in interpretation. But, I'm open to use the metrics that are sensible for both incident predictions and cumulative predictions. Feel free to add them in this issue if you find some!

PR #95, opened 2025-01-13
Pre-commit compliance

swo:
- Add some `NotImplementedError`s; I wasn't systematic about this
- Comment out code that doesn't actually function, but we want to use as a skeleton
- Ensure pre-commit passes


Fuhan-Yang (approved):
Looks good! Interestingly commenting the code didn't raise pre-commit error. I have the experiences that pre-commit complained about the unused commented code, not sure if it was black.

PR #98, opened 2025-01-15
Suggestion: update preprocessing

swo:
- Remove the concept of datasets
- Only return the cumulative uptake data
- Rename "rollout" to "rollout***s***" everywhere, to emphasize that you can insert multiple dates

### DIFF
--- a/scripts/preprocess.py
+++ b/scripts/preprocess.py
@@ -14,22 +14,20 @@ def preprocess(
     filters: dict,
     keep: List[str],
     groups: List[str],
-    rollout_dates: List[datetime.date],
-) -> pl.DataFrame:
+    rollouts: List[datetime.date],
+) -> iup.CumulativeUptakeData:
     # Prune data to correct rows and columns
-    cumulative_data = iup.CumulativeUptakeData(
-        raw_data.filter(**filters).select(keep).sort("time_end").collect()
-    )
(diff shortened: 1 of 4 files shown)


scripts/preprocess.py
```diff
@@ -14,29 +14,19 @@ def preprocess(
...
-    # Convert to incident data
-    incident_data = cumulative_data.to_incident(groups)
-
-    return pl.concat(
-        [
-            cumulative_data.with_columns(estimate_type=pl.lit("cumulative")),
-            incident_data.with_columns(estimate_type=pl.lit("incident")),
-        ]
+    # note the awkward wrapping with the class, because insert_rollouts returns
+    # a normal data frame
```
eschrom, line 27: Should we just change that? We can just convert the output to `CumulativeUptakeData` as the last step inside `insert_rollouts`, right?

eschrom (approved):
Looks good to me.

Fuhan-Yang (approved):
Thanks for clarifying in the yaml file that there is one data frame. This looks good to me!

scripts/preprocess.py
swo: Given the discussion elsewhere about using cumulative uptake for everything, I think it might be worth refactoring. I'll open an issue for this.

swo: See #99

PR #101, opened 2025-01-16
Several pipeline improvements

eschrom:
I ran the three scripts in the pipeline (preprocess.py, forecast.py, and eval.py) and found several things I wanted to improve:
- Several of the UptakeData methods are edited, along with their type-hinting, to return the correct UptakeData subclass, rather than generic polars data frames. This cleans up some of the excess type enforcing from the scripts.
- I clarified the way forecast dates are dealt with. Previously, every week between `["timeframe"]["start"]` and `["timeframe"]["end"]` in the config was used as a forecast date, resulting in multiple forecast trajectories of progressively shorter forecast horizons. Now, only the dates explicitly specified in `["timeframe"]["start"]` are used as forecast dates. If only one `["timeframe"]["start"]` is specified, then only one trajectory (per model) is forecast. If you want multiple forecast trajectories as the forecast date (i.e. `["timeframe"]["start"]`) gets pushed later into the season, shortening the forecast horizon, you must supply multiple `["timeframe"]["start"`] dates as a list in the config. The config_template.yaml is updated to show an example. This is useful, because now the forecast.py script can be used to either 1) explore model performance as the forecast date and horizon changes, or 2) make a single forecast trajectory from the last observed date into the future.
- Minor bug fixes and edits for conciseness.

### DIFF
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -110,9 +110,7 @@ class UptakeData(Data):
 
 
 class IncidentUptakeData(UptakeData):
-    def to_cumulative(
-        self, group_cols: List[str,] | None, last_cumulative=None
-    ) -> pl.DataFrame:
+    def to_cumulative(self, group_cols: List[str,] | None, last_cumulative=None):
         """
         Convert incident to cumulative uptake data.
 
@@ -145,7 +143,7 @@ class IncidentUptakeData(UptakeData):
                 estimate=pl.col("estimate") + pl.col("last_cumulative")
             ).drop("last_cumulative")
 
-        return out
+        return CumulativeUptakeData(out)
 
 
 class CumulativeUptakeData(UptakeData):
@@ -178,9 +176,7 @@ class CumulativeUptakeData(UptakeData):
 
         return IncidentUptakeData(out)
 
-    def insert_rollouts(
-        self, rollouts: List[dt.date], group_cols: List[str] | None
-    ) -> pl.DataFrame:
+    def insert_rollouts(self, rollouts: List[dt.date], group_cols: List[str] | None):
         """
         Insert into cumulative uptake data rows with 0 uptake on rollout dates.
 
@@ -228,7 +224,7 @@ class CumulativeUptakeData(UptakeData):
             == 0.0
         ).all()
 
-        return frame
+        return CumulativeUptakeData(frame)
 
 
 class QuantileForecast(Data):
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -45,7 +45,9 @@ class UptakeData(Data):
         self.assert_in_schema({"time_end": pl.Date, "estimate": pl.Float64})
 
     @classmethod
-    def split_train_test(cls, uptake_data, start_date: dt.date, side: str):
+    def split_train_test(
+        cls, uptake_data: "UptakeData", start_date: dt.date, side: str
+    ) -> "UptakeData":
         """
         Concatenate Uptake data objects and split into training and test data.
 
@@ -110,7 +112,9 @@ class UptakeData(Data):
 
 
 class IncidentUptakeData(UptakeData):
-    def to_cumulative(self, group_cols: List[str,] | None, last_cumulative=None):
+    def to_cumulative(
+        self, group_cols: List[str,] | None, last_cumulative=None
+    ) -> "CumulativeUptakeData":
         """
         Convert incident to cumulative uptake data.
 
@@ -176,7 +180,9 @@ class CumulativeUptakeData(UptakeData):
 
         return IncidentUptakeData(out)
 
-    def insert_rollouts(self, rollouts: List[dt.date], group_cols: List[str] | None):
+    def insert_rollouts(
+        self, rollouts: List[dt.date], group_cols: List[str] | None
+    ) -> "CumulativeUptakeData":
         """
         Insert into cumulative uptake data rows with 0 uptake on rollout dates.
 
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -1,5 +1,5 @@
 import datetime as dt
-from typing import List, Sequence
+from typing import List
 
 import polars as pl
 from polars.datatypes.classes import DataTypeClass
@@ -44,16 +44,14 @@ class UptakeData(Data):
         """
         self.assert_in_schema({"time_end": pl.Date, "estimate": pl.Float64})
 
-    @staticmethod
-    def split_train_test(
-        uptake_data_list: Sequence[Data], start_date: dt.date, side: str
-    ) -> pl.DataFrame:
+    @classmethod
+    def split_train_test(cls, uptake_data, start_date: dt.date, side: str):
         """
-        Concatenate Data objects and split into training and test data.
+        Concatenate Uptake data objects and split into training and test data.
 
         Parameters
-        uptake_data_list: Sequence[Data]
-            cumulative or incident uptake data objects, often from different seasons
+        uptake_data: UptakeData
+            cumulative or incident uptake data across all seasons
         start_date: dt.date
             the first date for which projections should be made
         side: str
@@ -65,23 +63,16 @@ class UptakeData(Data):
 
         Details
         Training data are before the start date; test data are on or after.
+        Infers what type of UptakeData to return from what type was given.
         """
         if side == "train":
-            out = (
-                pl.concat(uptake_data_list)
-                .sort("time_end")
-                .filter(pl.col("time_end") < start_date)
-            )
+            out = uptake_data.sort("time_end").filter(pl.col("time_end") < start_date)
         elif side == "test":
-            out = (
-                pl.concat(uptake_data_list)
-                .sort("time_end")
-                .filter(pl.col("time_end") >= start_date)
-            )
+            out = uptake_data.sort("time_end").filter(pl.col("time_end") >= start_date)
         else:
             raise RuntimeError(f"Unrecognized side '{side}'")
 
-        return out
+        return type(uptake_data)(out)
 
     @staticmethod
     def date_to_season(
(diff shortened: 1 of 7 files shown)


iup/__init__.py
```diff
@@ -65,23 +65,16 @@ def split_train_test(
...
-                pl.concat(uptake_data_list)
-                .sort("time_end")
-                .filter(pl.col("time_end") >= start_date)
-            )
+            out = uptake_data.sort("time_end").filter(pl.col("time_end") >= start_date)
         else:
             raise RuntimeError(f"Unrecognized side '{side}'")
 
-        return out
+        return type(uptake_data)(out)
```
Fuhan-Yang, line 77: This is a great way to provide an object with a class type, and get around the issue of using the class in arguments to return the same class. New method learned!

scripts/config_template.yaml
```diff
@@ -17,7 +17,7 @@ data:
 
 # The timeframe over which to generate projections
 timeframe:
-  start: 2024-02-01
+  start: [2024-02-03, 2024-02-10]
```
Fuhan-Yang, line 20: When evaluating the performance in several months even a year in the future, say separated by a week, this argument will be lengthy, and caution is needed to avoid incorrect dates.

Fuhan-Yang (commented):
Thanks Ed for debugging the pipeline! Whether using forecast start and end or a list of forecast dates to generate multiple trajectories depends on which option is more commonly used. 

1. Having a single forecast start and end is convenient for time-varying evaluation. Like if we want to know how the model performs say weekly in the next year, a single forecast start could generate a consistent evaluation, without having to input a list of 52 dates in the arguments (while we can also generate this list by coding and copy and paste, but the argument is just lengthy and not concise).

2. Having a list of forecast date is convenient if we only interested a single trajectory on a certain date. This is useful for a common task, forecasting the next entire season at the seasonal start. And we don't need to bother to extract this forecast from a list of trajectories with varied forecast dates within the season. 

That said, I can see the pros and cons for both design, and the decision should be made on which scenario is more common for the project. Nevertheless, if we want both, we can also develop an if-else option to control the program to generate a "single" forecast or "multiple" forecasts, given the forecast start and end.

iup/__init__.py
```diff
@@ -44,16 +44,16 @@ def validate(self):
...
         self.assert_in_schema({"time_end": pl.Date, "estimate": pl.Float64})
 
-    @staticmethod
+    @classmethod
     def split_train_test(
-        uptake_data_list: Sequence[Data], start_date: dt.date, side: str
-    ) -> pl.DataFrame:
+        cls, uptake_data: "UptakeData", start_date: dt.date, side: str
+    ) -> "UptakeData":
         """
```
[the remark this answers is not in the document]
eschrom: Yes! This is a great trick I learned from Chris. Usual type hinting (no quotes) doesn't work because this method belongs to the UptakeData class, so python hasn't yet understood what UptakeData is when it examines this method. Quotes tells python to wait until it has "read" the whole file before examining the type hinting. Check it out here: https://peps.python.org/pep-0484/#forward-references

scripts/config_template.yaml
eschrom: Yes, this argument could be lengthy. But I'm not sure how often we'll really want a list of dozens of dates in the `["timeframe"]["start"]` slot. I will make an illustration to explain.

eschrom:
In response to your overall comment:

Suppose we have data that looks like this:
image

In my approach, a single `["timeframe"]["start"]` and a single `["timeframe"]["end"]` will forecast a full trajectory with weekly predictions:
image

In your approach, a single `["timeframe"]["start"]` and a single `["timeframe"]["end"]` will forecast many trajectories with weekly predictions, each starting a week apart:
image

This works fine for the red curve, but now suppose we want to make projections for the blue curve.

My approach:
image

Your approach:
image

See how this doesn't make sense? If we are performing genuine forecasting (not simply evaluating models on past data), then your approach generates forecasts that start many weeks after the last observed data point.

Additionally, `["timeframe"]["interval"]` is the interval between predictions for a _single trajectory_. Your approach co-opts this config entry to also be the interval between dates on which multiple trajectories begin. Even if you are purely evaluating models on past data with multiple forecast horizons, what if you want each trajectory to contain weekly predictions, but you want each trajectory to begin 1 month apart?

All told, that's why I've restructured. I think my approach is more flexible to accommodate genuine forecasting as well as model evaluation, and with different forecast intervals vs. forecast horizons in the latter case.

Fuhan-Yang:
When we evaluate models, it should be based on the fact that the past data exists. And if we are performing genuine forecasts on partial season of data, we can extract the needed trajectory from the output of current model. Another reason I'm hesitant is that I'm not sure how frequent we need to manually generate different lists of dates, and if there is a clear way to record these arguments (maybe this is more reproducible in post-process analysis?). A single date is simpler and more clear as an argument. 

The issue can also be how much connected do we want for forecasting and evaluation functionality. Current structure combines these two together, like eval.py will always conduct on the results of forecast.py, given the same input in config. This is useful and common in retrospective forecasting, which we forecast on past data and evaluate in a time-varying manner. But you are right about there is situation that forecasting and evaluation aim for different inputs (like partial season of data, and different interval). We can solve this by designing separate inputs too. The package can choose to forecast only, or both in this case.

eschrom:
Indeed, a good long-term solution could be to separate the config inputs that describe the forecast trajectories from the config inputs that describe the evaluation forecast dates. What if we reformat that portion of the config like this:

forecast_timeframe          # Timeframe for the longest desired forecast
     start: 2024-02-03        # First day of the forecast trajectory
     end: 2024-04-30         # Last day of the forecast trajectory
     interval: 7d                  # Interval between forecasts along the trajectory

evaluation_timeframe      # Timeframe for evaluation
     interval: 1m                 # Interval between start dates for multiple forecast trajectories

This allows the interval between trajectory start dates (when performing evaluation) to differ from the interval between dates within a single forecast trajectory. Although you may be more interested in moving the trajectory start date 1 week at a time, I've used 1 month here just to show that it can be different. Moreover, if ["evaluation_timeframe"]["interval"] is left blank (i.e. 'None'), then `eval_all_forecasts` in the eval.py script could terminate itself, essentially skipping evaluation in the pipeline without having to design a totally separate pipeline for forecasting-only.

Does this sound good to you? If so, I'm happy to code it up and push to this PR.

Fuhan-Yang:
This sounds feasible to me! I like the idea to skip evaluation. In this case we can keep evaluation optional, and the package is more flexible in retrospective forecasting as well as real-time forecasting. Does this config mean that forecast.py will run a single trajectory based on the ["start"] and ["end"] of ["forecast_timeframe"], and the time-varying loop will be based on the same input, and will only come in when ["evaluation_timeframe"]["interval"] is not blank? Then we still have the same time inputs for both forecast and evaluation, but I think that's what we want. Thanks so much Ed! I will review the change!

eschrom:
Yes, your interpretation of my draft config is exactly correct. Fantastic! I will code and push the changes, then re-request your review. Thanks, Fuhan!

Fuhan-Yang (approved):
I think the change reflects exactly what we talked about. This looks good to me. Thank you so much, Ed!

swo:
I'm not sure I followed the full conversation here, but I want to re-iterate my terminology:

- The *observation date* is the date when the vaccination happens.
- The *forecast date* is when the forecast is run.
  - For prospective forecasts, this is today. You use all available data.
  - For retrospective forecasts, you use data that was available as of the forecast date (or, for simplicity, data with observation dates before the forecast date).
  - It doesn't really make sense to have a forecast date that is far into the future of the observation date of the most recent available data. Say you have data through day X, and run a forecast on day X. If you get no more data, and run a forecast on day X+1, it will be exactly the same as the forecast made on day X.
- I wonder if "timeframe" is a confusing word? I would write this as `first_forecast_date`, `last_forecast_date`, and `forecast_date_interval`.
- A *forecast* is a forecast date and a set of predictions (i.e., observation dates and predicted values on those dates). You can evaluate any forecast by comparing predictions to true data.

eschrom:
> * It doesn't really make sense to have a forecast date that is far into the future of the observation date of the most recent available data.

This is exactly the issue that we were discussing and have solved.

I find it odd to call each date for which a prediction is made an "observation date" in the prospective setting, because these dates are in the future so no observations have been made. But if that's the prevailing terminology, I'm happy to abide!

Using your terminology:

```
forecast_timeframe
     start: 2024-02-03    # The first forecast date, and the only one if forecasting is prospective
     end: 2024-04-30      # The last observation date of the forecast
     interval: 7d         # Interval between observation dates

evaluation_timeframe
     interval: 1m         # Interval between forecast dates if forecasting is retrospective. Leave blank if forecasting is prospective.
```

Fuhan-Yang:
We separated forecasting and evaluation to allow prospective forecast only. 

>       Say you have data through day X, and run a forecast on day X. If you get no more data, and run a forecast on day X+1, it will be exactly the same as the forecast made on day X.

 In this case, we can have prospective forecast by leaving ["evaluation_timeframe"]["interval"] blank. The forecast will start on the day X only, and generate a single forecast until ["forecast_timeframe"]["end"]. For example, we forecast the vaccine uptake from Jan 23 to Mar 30, given all data observed until Jan 22. 

>     A _forecast_ is a forecast date and a set of predictions (i.e., observation dates and predicted values on those dates). You can evaluate any forecast by comparing predictions to true data.

Do you mean retrospective forecast here? In this case, the forecast dates are within observation dates. By *not* leaving ["evaluation_timeframe"]["interval"] blank, we create a series of predictions iteratively with multiple forecast dates starting at ["forecast_timeframe"]["start"] and ending at ["forecast_timeframe"]["end"]. So we can evaluate them with true data. An example is: we observed vaccine uptake from Oct 1 2024 to Jan 22 2025, we start to forecast on every Monday during this time, and evaluate against observed data.

swo:
*Prospective* forecasting means making predictions for dates that do not have data yet. If you did a forecast today, 2025-01-22, about what uptake would be in February 2025, that is prospective.

All the forecasts we're doing right now in this project are *retrospective*. We're pretending we're in the past, and artificially limit what data we have available, to make predictions about dates for which we actually have data but just pretend we don't, for purposes of evaluating the models.

PR #102, opened 2025-01-23
Transform the linear incident uptake model to numpyro

eschrom:
The linear incident uptake model (the only structure we currently have) is now implemented in numpyro. Here are some improvements yet to come, but perhaps outside the scope of this PR:
- The priors for this model are somewhat foolish. It will take some work to choose smarter priors.
- The output of the model is just a polars `DataFrame`, not a `SampleForecast` object, because the samples are given in "wide" format (one column per sample) rather than "long" format (one column for sample ID and one column for the sample itself). This doesn't really matter yet, since no evaluation metrics use the prediction credible intervals. Samples are merely averaged and then discarded at the start of the `eval.py` script.
- Some tests for the numpyro model do not work and are commented out. How to test well when using numpyro is an ongoing exploration.


swo:
> * The output of the model is just a polars `DataFrame`, not a `SampleForecast` object, because the samples are given in "wide" format (one column per sample) rather than "long" format (one column for sample ID and one column for the sample itself).

Is this a common feature of numpyro outputs?

My instinct is that we should cast to "long" format, since that will make later parsing easier, and will also improve (I think) compression with parquet

iup/models.py
```diff
@@ -36,9 +39,44 @@ class LinearIncidentUptakeModel(UptakeModel):
...
+        daily: numpy array
+            standardized values of daily average incident uptake, the outcome
+
+        Returns
+        Nothing
+
+        Details
+        Provides the model structure and priors for a linear incident uptake model.
+        """
+        a = numpyro.sample("a", dist.Normal(0.0, 1.0))
```
swo, line 66: I think best practice here would be to say that the SD on `a` is a hyperparameter, so rather than hard-coding `1.0` you would have some argument to this function. This would let you play with priors by adjusting the config, rather than adjusting the code.

Although I'm seeing that this would probably require changing the config to take not just model names, but combination of model names and parameters:

```yaml
models:
    - name: LinearIncidentUptakeModel
      params:
        - a_SD: 1.0
        - bP_SD: 1.0
```

Then you could, say, do forecasts using the same model but with different priors, all in the same run, and then ask whether not just the selection of model but also the selection of priors for a model improves performance.

This is probably out of scope but I think a good idea at some point.

iup/models.py
```diff
@@ -151,16 +187,15 @@ def fit(self, data: IncidentUptakeData, group_cols: List[str,] | None) -> Self:
 
         self.standards = self.extract_standards(data, ("previous", "elapsed", "daily"))
 
-        self.x = (
-            data.select(["previous_std", "elapsed_std"])
-            .with_columns(interact=pl.col("previous_std") * pl.col("elapsed_std"))
-            .to_numpy()
+        self.kernel = NUTS(self.model)
+        self.mcmc = MCMC(self.kernel, num_warmup=1000, num_samples=2000)
```
swo, line 191: I would think that you could change the number of warmup and samples here, turn those into adjustable parameters (read from argument inputs), so that you could do a unit test that ran only a few steps?

cc @afmagee42 because I know you and @eschrom have talked a lot about this and I'm ignorant yet

iup/models.py
```diff
@@ -380,55 +415,63 @@ def project_sequentially(
...
-            x = np.insert(x, 2, np.array((x[:, 0] * x[:, 1])), axis=1)
-            # Predict the uptake on the projection date
-            y = model.predict(x)
+
+            # Predict the uptake on the projection date:
+            # One prediction per MCMC sample (row) for each prev*elap combo (col)
+            y = predictive(rng_key, previous=prev, elapsed=elap)["obs"]
+
+            # Randomly choose a prediction for each prev*elap combo
+            idx = np.random.randint(0, y.shape[0] - 1, size=y.shape[1])
```
swo, line 453: I think the more pythonic thing here would be to use https://numpy.org/doc/2.1/reference/random/generated/numpy.random.Generator.choice.html

But also, we'll want to track the seed that is used for this RNG also. Numpy now recommends using https://numpy.org/doc/stable/reference/random/generator.html objects, rather than setting a global seed.

Presumably we could do something like use the same integer as the starting seed for both the JAX RNG key as well as for the numpy seed?

Or is this something that should be left outside the model? Like, does "prediction" in a sample-based framework mean "give me *all* the posterior trajectories?"

Curious @afmagee42 what you do with the `linmod` stuff. Does `predict()` return a single trajectory? Many? An adjustable number?

iup/models.py
```diff
@@ -380,55 +415,63 @@ def project_sequentially(
...
                 standards["daily"]["std"],
             )
-            # This projection becomes the previous projection
-            # in the next loop iteration.
+            # This projection becomes 'previous' in the next loop iteration.
 
-        return proj
+        # Sequential projections become rows and sample trajectories cols.
+        # Remove the first row, which is the last known uptake value.
+        return np.delete(proj.transpose(), 0, 0)
```
swo, line 466: Low priority: Is there a way to write this as a slice, rather than a deletion?

Like I'm not sure if `proj.transpose()[1:, 1:]` does the trick

I had to go and look up `np.delete` to assure myself it wasn't one of the in-place operations, which I would have found confusing in a return statement

iup/models.py
```diff
@@ -380,55 +415,63 @@ def project_sequentially(
...
 
     def predict(
         self,
         start_date: dt.date,
         end_date: dt.date,
         interval: str,
         group_cols: List[str,] | None,
-    ) -> CumulativeUptakeData:
+    ) -> pl.DataFrame:
```
swo, line 474: See my comment in the main convo tab re: wide vs. long

scripts/eval.py
```diff
@@ -66,5 +66,8 @@ def eval_all_forecasts(data, pred, config):
     pred = pl.scan_parquet(args.pred).collect()
     data = pl.scan_parquet(args.obs).collect()
 
+    # Drop all samples and just use mean estimate, for now
+    pred = pred.drop([col for col in pred.columns if "estimate_" in col])
```
swo, line 70: Low priority I think the polaronic way to do this is:

```python
import polars.selectors as cs
pred = pred.select(~cs.starts_with("estimate_"))
```

https://docs.pola.rs/api/python/stable/reference/selectors.html

scripts/forecast.py
```diff
@@ -76,6 +76,17 @@ def run_forecast(
         grouping_factors,
     )
 
+    cumulative_projections = cumulative_projections.with_columns(
+        pl.Series(
+            "estimate",
+            cumulative_projections.select(
+                [col for col in cumulative_projections.columns if "estimate_" in col]
```
swo, line 83: And then this could be:
```python
cumulative_projections.select(cs.starts_with("estimate_"))
```

scripts/forecast.py
```diff
@@ -76,6 +76,17 @@ def run_forecast(
...
+            )
+            .to_numpy()
+            .mean(1),
```
swo, line 86:
```suggestion
            )
            .mean()
            .to_numpy()
```
Personally I find it easier to think in polars-land, rather than in both polars and in numpy

tests/test_linear_incident_uptake_model.py
```diff
@@ -45,6 +43,25 @@ def frame():
     return frame
 
 
+# This doesn't work, because passing 'model' to a test causes 'model' to be executed!
```
swo, line 46: Let's migrate this commented material into a new PR

I'd rather have a draft PR with an in-progress thought, rather than commented code committed to main

tests/test_linear_incident_uptake_model.py
```diff
@@ -150,30 +169,34 @@ def test_build_scaffold_handles_groups():
...
-
-    output = iup.models.LinearIncidentUptakeModel.project_sequentially(
-        elapsed, start, standards, model
-    )
-
-    assert all(
-        np.round(np.delete(output, 0), decimals=10)
-        == np.round(np.array(elapsed), decimals=10)
-    )
+# This doesn't work because passing 'model' to this test causes 'model' to be executed!
```
swo, line 172: ditto

swo (changes requested):
Some clarifying & organizing things. Feel free to punt things to downstream issues or ask me to do some refactoring

afmagee42:
> > * The output of the model is just a polars `DataFrame`, not a `SampleForecast` object, because the samples are given in "wide" format (one column per sample) rather than "long" format (one column for sample ID and one column for the sample itself).
> 
> Is this a common feature of numpyro outputs?
> 
> My instinct is that we should cast to "long" format, since that will make later parsing easier, and will also improve (I think) compression with parquet

To be clear, NumPyro doesn't touch dataframes. It [returns a `dict`](https://num.pyro.ai/en/stable/mcmc.html#numpyro.infer.mcmc.MCMC.get_samples) with the quantities you've specified to log (everything declared in a `numpyro.sample()` or `numpyro.deterministic()`). I find its structure... annoying at best.

Everything past that dict is up to you, the user. STF likes ArviZ, which uses a [different (seemingly less annoying)](https://python.arviz.org/en/stable/getting_started/XarrayforArviZ.html) format, and has a [NumPyro interface](https://python.arviz.org/en/stable/getting_started/CreatingInferenceData.html#from-numpyro).

iup/models.py
afmagee42: > Or is this something that should be left outside the model? Like, does "prediction" in a sample-based framework mean "give me _all_ the posterior trajectories?"

In a Bayesian context, in general we get _distributions_ not values. That holds with predictions, where we get predictive distributions. Condensing that down to a single summary is a whole thing. But it's a pretty modular decision and should be pushed as close to where it's needed as possible and not be done before.

There are packages that actually return _more_ samples from the posterior predictive distribution than posterior samples because they sample multiple times from the likelihood per posterior sample for smoothness. 

Other than thinning to use only some fraction of samples where storage becomes a problem, I know of nothing that returns fewer predictive distribution samples than posterior distribution samples.

There is a sense in which any one posterior sample is as good of a single-number summary of the whole distribution as any other. But posterior means and medians are more common choices, and unless we want to be in the business of offering a bunch of options here (we don't), we should modularize that decision and return the distribution. Then let the user decide later when they need that one summary what it should be.

> Curious @afmagee42 what you do with the `linmod` stuff. Does `predict()` return a single trajectory? Many? An adjustable number?

In linmod, we work with distributions until we absolutely have to summarize them. We don't currently have an explicit `Model.predict()`, but when we need predictions on observables we use the whole [kit and caboodle](https://github.com/CDCgov/cfa-viral-lineage-model/blob/35fc92e6619513c11af9afa275c5a7bf1fcac3a2/linmod/eval.py#L173).

> Presumably we could do something like use the same integer as the starting seed for both the JAX RNG key as well as for the numpy seed?

Funny enough, linmod actually also has numpy RNGs in it for generating predictions anyways, though. It's how we sample from the multinomial. My take is it doesn't really matter how you handle the seeds behind the scene as long as a user only has to set one. But the easiest thing to do is to use the same integer for both seeds.

iup/models.py
```diff
@@ -36,9 +39,44 @@ class LinearIncidentUptakeModel(UptakeModel):
 
     def __init__(self):
         """
-        Initialize the model as a scikit-learn linear regression.
+        Initialize with a seed and the model structure.
         """
-        self.model = LinearRegression()
+        self.rng_key = random.PRNGKey(0)
```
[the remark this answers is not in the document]
eschrom: The seed is required to fit the model, so it would have to be an argument of `LinearIncidentUptakeModel.fit()` upstream of `LinearIncidentUptakeModel.predict()`, too. But regardless, the point is that the seed should be in the config for easy manipulation, right?

iup/models.py
```diff
@@ -36,9 +39,44 @@ class LinearIncidentUptakeModel(UptakeModel):
...
+        Nothing
+
+        Details
+        Provides the model structure and priors for a linear incident uptake model.
+        """
+        a = numpyro.sample("a", dist.Normal(0.0, 1.0))
+        P, E, PE = 0.0, 0.0, 0.0
+        if previous is not None:
+            bP = numpyro.sample("bP", dist.Normal(0.0, 1.0))
+            P = bP * previous
```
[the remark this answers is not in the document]
eschrom: Yes, these all look equivalent to me.

iup/models.py
eschrom: Yes, this is exactly what I want to work toward! And I totally agree that the config will require these new inputs. First, however, I want to do some informal exploration of what the right shapes are for the priors. E.g. non-Gaussian priors might prevent incident uptake estimates from being absurdly negative for long forecast horizons. With any luck, the prior shapes can be hard-coded, but the exact parameters can be tuned in the config as you've sketched. See #104

iup/models.py
eschrom: I agree, this is also my best guess at how the model can be tested efficiently (as opposed to some pseudo-model fixture, which is what I tried and failed to do in this PR). Again, see #104.

iup/models.py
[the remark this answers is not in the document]
afmagee42: Config is ideal. In linmod it's hard-coded it in the pipeline, which is less ideal but still configurable without touching core code https://github.com/CDCgov/cfa-viral-lineage-model/blob/35fc92e6619513c11af9afa275c5a7bf1fcac3a2/retrospective-forecasting/main.py#L69

iup/models.py
eschrom: I'm happy to update the way a random number is picked.

I don't think it's feasible to get _all_ the posterior trajectories here. Let me clarify how trajectories are made:

The model predicts next week's incident uptake from this week's incident uptake. Suppose you have 1000 posterior samples of the model's parameters. Then (without the thinning or oversampling that Andy mentioned), just one value for this week's uptake will yield 1000 estimates of next week's uptake.

To predict a full trajectory, prediction must be iterated sequentially: Week 1 predicts Week 2, which predicts Week 3, which predicts Week 4, and so on.
- If you know Week 1's uptake, you get 1000 estimates of Week 2's uptake.
- If you have 1000 estimates of Week 2's uptake, and each produces 1000 estimates of Week 3's uptake, you get 1,000,000 estimates of Week 3's uptake.
- The number of samples grows exponentially as you move into the future.

Alternatively, you could imagine this workflow:
- If you know Week 1's uptake, you get 1000 estimates of Week 2's uptake.
- Take just the median of Week 2's estimates and get 1000 estimates of Week 3's uptake.
- The number of samples stays constant as you move into the future.
- BUT, this does not appropriately compound uncertainty to make a cone as you move into the future.

So my solution is:
- If you know Week 1's uptake, you get 1000 estimates of Week 2's uptake.
- For each Week 2 estimate, randomly select one of the 1000 estimates of Week 3's uptake.
- The number of samples stays constant as you move into the future.
- Uncertainty does compound as you move into the future.

iup/models.py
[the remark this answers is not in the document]
afmagee42: Scott the last one of yours guarantees an extra variable lying around. This is good if NumPyro gets tetchy about model dimensions being consistent. But I don't think it does, so it's bad because you're needlessly increasing model dimension.

iup/models.py
eschrom: Wide is not any sort of numpyro default - it's just what came naturally to me. If we're confident long is better, I'm happy to change it.

scripts/eval.py
eschrom: Looks like a useful trick - I would happily get used to this shorter syntax.

iup/models.py
afmagee42: I don't see how making them flexible is a problem here? Can we not simply change

```python
def fit(self, data: IncidentUptakeData, group_cols: List[str,] | None) -> Self:
    ...
    self.mcmc = MCMC(self.kernel, num_warmup=1000, num_samples=2000)
```

to
```python
def fit(self, data: IncidentUptakeData, group_cols: List[str,] | None, num_warmup: int=1000, num_samples: int=2000) -> Self:
    ...
    self.mcmc = MCMC(self.kernel, num_warmup=num_warmup, num_samples=num_samples)
```

We could make it even more configurable with `**kwargs` but that might be overkill.

afmagee42: I also would like to urge us to use _multiple chains_. While we can never be sure MCMC has converged, having multiple independent chains sampling the same thing makes me a heck of a lot more comfortable than one chain.

[stan defaults to 4](https://pystan.readthedocs.io/en/latest/reference.html)

iup/models.py
afmagee42: Let's set aside time for a synchronous chat about this, and defer it to a later PR. I am, perhaps overly, confident that there is a way to either rewrite this model, or write a sufficiently similar one, that fits into more normal modeling setups and where it's perfectly straightforward to sample one curve from the prior.

afmagee42: Also, if we have a writeup of the model in probability distribution form, that would be rather helpful.

Like, if $\upsilon\_t$ is the unobserved true latent uptake at time t, I'd like to see specification of $\text{Pr}(\upsilon\_{t + 1} \mid \upsilon\_t, ..., \upsilon\_0)$.

Semi-separated stochastic and deterministic components would be okay too, $\upsilon\_{t + 1} = f(\upsilon\_t, ..., \upsilon\_0) + \xi\_t$, where $\xi\_t \sim \text{someDistribution}$.

eschrom: Sure, let $\upsilon_t$ be the incident uptake during week $t$, where $t$ is the number of weeks elapsed since rollout.

Then the model is: 
$\upsilon_{t+1} \sim Normal(\mu, \sigma)$
$\mu = \alpha + \beta_{\upsilon}\upsilon_t + \beta_{t}t + \beta_{\upsilon*t}\upsilon_{t}t$

where priors are:
$\alpha \sim \text{someDistribution}$
$\beta_{\upsilon} \sim \text{someDistribution}$
$\beta_{t} \sim \text{someDistribution}$
$\beta_{\upsilon*t} \sim \text{someDistribution}$
$\sigma \sim \text{someDistribution}$

afmagee42: Thanks, Ed! Would you mind if I play around with reworking things a bit and maybe open up a separate PR?

eschrom: Of course! If you have the bandwidth, go for it.

swo: Let's open an issue for this, to dig into

I read this function as drawing a single trajectory rather than, as @afmagee42 points out, a distribution on posterior predictions (for some date[s]). Does that mean that we should draw a single `idx`, which is the index of a single posterior sample (particle?), and generate the trajectory from that? (Not sure any of the clauses in that sentence are necessarily true. But it feels weird that we need to draw a new `idx` for each `i`.)

swo: Setting up future issue as https://github.com/CDCgov/cfa-immunization-uptake-projection/issues/106

swo:
> But I would tend to think long-formatting posteriors is a bit insane. You repeat the sample ID over and over and over for all logged parameters that way.

I think this doesn't matter for a columnar, compressed data format like parquet: it will do run-length encoding and other tricks that make the repeated entries no problem.

Long/tidy format is so much easier to work with as a user, that I'd want to see some evidence that it's substantially worse in time or space (i.e., time to load/save, or space on disk, or in memory) to make it worth doing wide

iup/models.py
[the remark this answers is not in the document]
eschrom: Forthcoming updates will have the seed as a model parameter in the config.

iup/models.py
eschrom: Forthcoming updates will have prior distribution parameters be adjustable from the config.

iup/models.py
eschrom: Forthcoming updates will have mcmc control parameters be adjustable from the config, including multiple chains.

iup/models.py
eschrom: `proj.transpose()[1:, :]` is what we want. This will be changed in forthcoming updates.

scripts/eval.py
eschrom: This will be moot if we switch to long format.

scripts/forecast.py
eschrom: This will be moot if we switch to long format.

scripts/forecast.py
eschrom: Forthcoming updates switch to long format.

scripts/forecast.py
eschrom: Forthcoming updates switch to long format.

scripts/eval.py
eschrom: Forthcoming updates switch to long format.

iup/models.py
eschrom: Forthcoming updates switch to wide format.

tests/test_linear_incident_uptake_model.py
eschrom: Forthcoming updates will fix this, now that mcmc control parameters can be adjusted from the config.

tests/test_linear_incident_uptake_model.py
eschrom: Forthcoming updates will fix this, now that mcmc control parameters can be adjusted from the config.

iup/models.py
eschrom: @afmagee42 and I have discussed, and I am now clear on what must be fixed. Thank you, Andy! More updates will be forthcoming today to correct `project_sequentially` accordingly.

Fuhan-Yang: I want to make sure that I totally understand: the posterior distributions of parameter estimates ($\beta_{\upsilon}$, $\beta_{t}$ and $\beta_{\upsilon*t}$ ) are used to generate posterior predictive distribution  $v_{t}$. But in our iterative prediction structure, how to select $v_{t}$ to represent as a predictor to predict the next $v_{t+1}$ remains a question. Currently we scrambled the indices of the posterior prediction distribution to combine with another predictor `elapsed`

Fuhan-Yang (approved):
Great work, Ed! Learning numpyro and implementing it to our current infrastructure is substantial work. My comments are checking my understanding, as I'm learning this I haven't had many constructive suggestions. This looks good, I think the next step naturally goes to adding relevant metrics, like QS, WIS which evaluate distributions. Maybe this will be my ticket in next sprint. But overall, great job to take our project into a Bayesian world!

eschrom:
@Fuhan-Yang - I have caused some confusion over how forward projections are generated from a fit `LinearIncidentUptakeModel`, so I will try to clarify. With @afmagee42's help, I have changed the code since you reviewed it, with the randomization step removed.

Let $\upsilon_t$ be the incident uptake during week $t$, where $t$ is the number of weeks elapsed since rollout.

Then the model is: 
$\upsilon_{t+1} \sim Normal(\mu, \sigma)$
$\mu = \alpha + \beta_{\upsilon}\upsilon_t + \beta_{t}t + \beta_{\upsilon*t}\upsilon_{t}t$

where priors are:
$\alpha \sim \text{someDistribution}$
$\beta_{\upsilon} \sim \text{someDistribution}$
$\beta_{t} \sim \text{someDistribution}$
$\beta_{\upsilon*t} \sim \text{someDistribution}$
$\sigma \sim \text{someDistribution}$

Once the model is fit, we have D draws from the posterior, where each draw is of the form $(\alpha, \beta_t, \beta_{\upsilon}, \beta_{\upsilon*t}, \sigma)$. Each draw should be used to simulate an entire forward trajectory. 

In more detail, if $T$ is the last observed reference date, then Draw 1 is used to compute $\upsilon_{T+1}$, which will be a distribution with standard deviation $\sigma$. One value is sampled from the $\upsilon_{T+1}$ distribution and used as the predictor to obtain $\upsilon_{T+2}$. This repeats forward in time, to obtain one full trajectory. The whole process is repeated for Draw 2, ... D, to obtain D trajectories, which together give our forecast and its uncertainty.

iup/models.py
```diff
@@ -103,7 +160,13 @@ def extract_standards(data: IncidentUptakeData, var_cols: tuple) -> dict:
...
 
-    def fit(self, data: IncidentUptakeData, group_cols: List[str,] | None) -> Self:
+    def fit(
+        self,
+        data: IncidentUptakeData,
+        group_cols: List[str,] | None,
+        params: dict,
+        mcmc: dict,
+    ) -> Self:
         """
```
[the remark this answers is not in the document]
swo: I think of the `return self` line as just convenience. You could leave that off and it would work basically the same

The question about re-fitting gets to the bit about keys in my note above. In my ideal world, there is a deterministic link between input parameters and output values, and so that ultimately means that the RNGs for the MCMC fitting and the trajectory drawing need to be considered as input parameters

iup/models.py
```diff
@@ -150,6 +175,10 @@ def fit(self, data: IncidentUptakeData, group_cols: List[str,] | None) -> Self:
             training data on which to fit the model
         group_cols: (str,) | None
             name(s) of the columns for the grouping factors
+        params: dict
```
swo, line 178: It's a bit funny that `lium()` takes the parameters as individual arguments, but this takes a dictionary

I could imagine that you make the hyperparameters a nested dictionary:

```python
params = {"hyperparams": {"a_mn": 0.0, etc}, etc}
```

So that you could say `self.mcmc.run(rngkey, previous, elapsed, daily, **params["hyperparams"])

iup/models.py
```diff
@@ -445,13 +486,11 @@ def project_sequentially(
                 proj.shape[0],
             )
 
-            # Predict the uptake on the projection date:
+            # Predict the uptake on the next date:
```
swo, line 489: Cf. discussions in other venues; I think we'll settle on some common nomenclature for this stuff

scripts/config_template.yaml
```diff
@@ -25,7 +25,25 @@ forecast_timeframe:
...
+      bE_mn: 0.0,
+      bE_sd: 1.0,
+      bPE_mn: 0.0,
+      bPE_sd: 1.0,
+      sig_mn: 1.0},
+    mcmc:
+      {num_warmup: 1000,
+      num_samples: 1000,
+      num_chains: 4}
+  }]
```
swo, line 46:
```suggestion
  - name: LinearIncidentUptakeModel
    seed: 0
    params:
      a_mn: 0.0
      a_sd: 1.0
      bP_mn: 0.0
      bP_sd: 1.0
      bE_mn: 0.0
      bE_sd: 1.0
      bPE_mn: 0.0
      bPE_sd: 1.0
      sig_mn: 1.0
    mcmc:
      num_warmup: 1000
      num_samples: 1000
      num_chains: 4
```
In yaml, you can use the `-` to indicate that that's a list item, but then the list item can just be a dictionary (with nested dictionaries)

scripts/forecast.py
```diff
@@ -66,8 +63,17 @@ def run_forecast(
...
     )
 
-    # Fit models using the training data and make projections
-    fit_model = model().fit(incident_train_data, grouping_factors)
+    # Make an instance of the model, fit it using training data, and make projections
+    assert issubclass(getattr(iup.models, model["name"]), iup.models.UptakeModel), (
+        f"{model['name']} is not a valid model type!"
+    )
+    ""
+    fit_model = getattr(iup.models, model["name"])(model["seed"]).fit(
```
swo, line 71:
```suggestion
    model_class = getattr(iup.models, model["name"])
    assert issubclass(model_class), iup.models.UptakeModel), (
        f"{model['name']} is not a valid model type!"
    )
    ""
    fit_model = model_class(model["seed"]).fit(
```

scripts/forecast.py
```diff
@@ -76,15 +82,10 @@ def run_forecast(
...
-            cumulative_projections.select(
-                [col for col in cumulative_projections.columns if "estimate_" in col]
-            )
-            .to_numpy()
-            .mean(1),
-        )
+    cumulative_projections = (
+        cumulative_projections.group_by(config["data"]["groups"] + ["time_end"])
+        .agg(pl.col("estimate").mean().alias("estimate"))
+        .sort("time_end")
```
swo, line 88: The alternative to sorting by time_end is to `.group_by(preserve_order=True)`. But I think this is probably less ambiguous/prone to error

tests/test_linear_incident_uptake_model.py
```diff
@@ -43,23 +43,35 @@ def frame():
...
+        "bPE_mn": 0.0,
+        "bPE_sd": 1.0,
+        "sig_mn": 1.0,
+    }
+
+    return params
+
+
+@pytest.fixture
+def mcmc():
```
swo, line 68: This naming is a little confusing because `mcmc` is used above to refer to the object produced by `MCMC()`. So maybe this should be `mcmc_params` and the thing above is `hyperparams` or something?

swo (approved):
I'm out of my depth on numpyro models, so I just have some comments about nomenclature, design, etc. that we can mostly think about later.

iup/models.py
```diff
@@ -37,15 +37,32 @@ class LinearIncidentUptakeModel(UptakeModel):
     - interaction of "elapsed" and "previous"
     """
 
-    def __init__(self):
+    def __init__(self, seed: int):
```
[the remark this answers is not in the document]
swo: Created as #110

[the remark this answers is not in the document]
eschrom: Happy to punt this to an issue while I learn more about it.

iup/models.py
eschrom: My thought process is that `lium()` always knows exactly which parameters it wants. In contrast, `fit()` may find itself fitting any of several pre-defined models, which each want different parameters.

scripts/config_template.yaml
eschrom: I like this! I didn't understand what it was the first time I saw it. I will make the change.

scripts/forecast.py
eschrom: This looks reasonable to me, but it upsets pylance.

scripts/forecast.py
eschrom: My thought process exactly.

iup/models.py
swo: Great point; this means we have a common signature for `fit()`, which is good

scripts/forecast.py
swo: weird

PR #103, opened 2025-01-23
Use snakemake

swo:
I hope that this PR can:

1. Show what snakemake can do, especially around separating jobs
2. Clarify what "forecast dates" and "target dates" are

Content of the PR:

- Create a [Snakefile](https://snakemake.readthedocs.io/en/stable/snakefiles/rules.html)
- Revamp the workflow so that:
  - The data are cached all in one step
  - The data are preprocessed all in one step
  - Each forecast (i.e., combination of model and forecast date) is run as a separate job
    - This requires revamping `forecast.py`
    - :warning: This does not yet account for grouping factors
    - :warning: This does not yet account for which target dates each forecast should include
  - Outputs of each forecast are kept in a separate parquet, but in [Hive format](https://docs.pola.rs/user-guide/io/hive/) so that they can all be opened and scanned in one go
  - Run all the scoring at once, since this should be cheap

Also:

- Update python version
  - This was necessary because snakemake requires Python >=3.11
  - This will require some updating of python versions by other users, if we end up wanting to use this
- Add snakemake as a dependency


swo:
@eschrom @Fuhan-Yang This isn't at all ready to be merged in, so I'm leaving it as a draft PR, but I would appreciate your eyes on this.

1. Does this make sense, this way of structuring the jobs?
2. Does it seem like snakemake is worth the effort to implement? Personally I *love* the idea of having a separate job per model+forecast date
3. Clarifications, questions, etc. etc. as always!

swo:
As per discussion with @eschrom , this is premature. Let's wait until we feel like it's too slow to run all the models & forecast dates, and then revisit this

PR #105, opened 2025-01-25
Draft analytical plan

swo:
Not sure if this is the best place for it, but I wanted to double-check we are on the same page about how to proceed.

The term "reference date" is what I had previously called "observation date." The terminology "reference date" is what other CFA teams use; we should move to that for consistency. (Thanks @zsusswein for educating me on this!)

### DIFF
--- a/docs/analytical_plan.md
+++ b/docs/analytical_plan.md
@@ -1,0 +1,34 @@
+# Draft analytical plan
+
+## Definitions
+
+Every data point has an associated _reference date_, when the measured physical process occurred, and a _report date_, when that data became available for forecasting. For example, if immunization uptake was 50% on April 1, and that information became available on April 5, the reference date is April 1 and the report date is April 5.
+
+A _forecast_ is:
+
+- a model, including its parameters and training data,
+- a _forecast date_,
+- a set of reference dates after the future relative to the forecast date, and
+- the predicted values for each reference date.
+
+In _prospective_ forecasting, the forecast date is the moment that the forecast was actually run. The training data is all the data that is available as of that moment, and the predictions are for reference dates after the forecast date.
+
+In _retrospective_ forecasting, the forecast date is a date in the past (relative to today). The training data is all the data that was available as of the forecast date, that is, all the data that has a report date before the forecast date. Predictions are for reference dates after the forecast date but before today.
+
+The forecast _horizon_ is the difference between the forecast date and a forecast reference date. A single forecast can cover multiple horizons. For example, a forecast with a forecast date of October 1 might include forecasts for the reference dates October 14, October 21, and so forth.
+
+A _score function_ takes predicted values produced by a single model (usually at some specific horizon), compares them to the true values, and returns a _score_ that quantifies the predictive performance of that model.
+
+## Analytical plan
+
+Perform retrospective forecasting to compare model performance:
+
+- Define the _start of season_ as September 1(?) and _end of season_ as April 1(?) or similar, TBD
+- For the 3(?) most recent seasons, let the forecast dates be every Saturday from the start of the season to the end of the season
+- For each model, forecast season, and forecast date: predict uptake for each week from the forecast date to the end of the season
+  - Note that this includes training data from prior seasons as well as that season up to the forecast date
+  - Note that earlier forecasts will produce more predictions (and horizons) than later forecasts
+- For each model, score function, and horizon: generate a score
+  - For the end-of-season uptake score function(s), score(s) will evaluate predictions for a single reference date (for each horizon)
+  - For other score functions, scores will include multiple reference dates for each horizon
+- Compare the scores across models (stratifying by horizon) and within models (across horizons)


docs/analytical_plan.md
```diff
@@ -0,0 +1,34 @@
...
+
+## Definitions
+
+Every data point has an associated _reference date_, when the measured physical process occurred, and a _report date_, when that data became available for forecasting. For example, if immunization uptake was 50% on April 1, and that information became available on April 5, the reference date is April 1 and the report date is April 5.
+
+A _forecast_ is:
+
+- a model, including its parameters and training data,
+- a _forecast date_,
+- a set of reference dates after the future relative to the forecast date, and
```
eschrom, line 11: I find "after the future" confusing. I would rephrase this to either:
"- a set of references dates in the future relative to the forecast date"
or
"- a set of reference dates occurring after the forecast date"

docs/analytical_plan.md
```diff
@@ -0,0 +1,34 @@
...
+- the predicted values for each reference date.
+
+In _prospective_ forecasting, the forecast date is the moment that the forecast was actually run. The training data is all the data that is available as of that moment, and the predictions are for reference dates after the forecast date.
+
+In _retrospective_ forecasting, the forecast date is a date in the past (relative to today). The training data is all the data that was available as of the forecast date, that is, all the data that has a report date before the forecast date. Predictions are for reference dates after the forecast date but before today.
```
eschrom, line 16: This seems more detailed than what we actually do. Report dates are not recorded in the NIS data, only reference dates, right? That is, if we made a retrospective forecast with forecast date Feb 1 2023, we would include the reference date Jan 31 2023 in the training data. We would not exclude the reference date of Jan 31 2023 because we assume that its report date could have been Feb 4 2023, after the forecast date.

While your definition makes sense in principle, it seems that in practice, "the training data is all the data with a reference date before the forecast date."

We probably need not split hairs over this - just thought it was worth mentioning.

docs/analytical_plan.md
```diff
@@ -0,0 +1,34 @@
...
+
+- Define the _start of season_ as September 1(?) and _end of season_ as April 1(?) or similar, TBD
+- For the 3(?) most recent seasons, let the forecast dates be every Saturday from the start of the season to the end of the season
+- For each model, forecast season, and forecast date: predict uptake for each week from the forecast date to the end of the season
+  - Note that this includes training data from prior seasons as well as that season up to the forecast date
+  - Note that earlier forecasts will produce more predictions (and horizons) than later forecasts
+- For each model, score function, and horizon: generate a score
+  - For the end-of-season uptake score function(s), score(s) will evaluate predictions for a single reference date (for each horizon)
+  - For other score functions, scores will include multiple reference dates for each horizon
+- Compare the scores across models (stratifying by horizon) and within models (across horizons)
```
eschrom, line 34: Should performing prospective forecasting come September 2025 be part of our analytical plan, too?

eschrom (approved):
This is _super_ helpful. I wouldn't even mind having this in the README, but a separate doc is fine too. Just a few questions, peppered throughout.

docs/analytical_plan.md
Fuhan-Yang: I'm not sure if the dates in the NIS data are already report dates? Think about the data on Jan 31 2023, maybe the actual survey was already done on Jan 26 2023, and took five days to reflect in NIS data. If this is the case, we should rephrase as "Predictions are for report dates after the forecast date but before today." It will be helpful to know when to use reference date or report date.

docs/analytical_plan.md
```diff
@@ -0,0 +1,34 @@
+# Draft analytical plan
+
+## Definitions
+
+Every data point has an associated _reference date_, when the measured physical process occurred, and a _report date_, when that data became available for forecasting. For example, if immunization uptake was 50% on April 1, and that information became available on April 5, the reference date is April 1 and the report date is April 5.
+
```
Fuhan-Yang, line 6: In forecasting, when is report date used? Is it a specific term when talking about report delay?

Fuhan-Yang (commented):
Thanks for clarifying the terminologies! My questions are about the user case of reference date versus report date. Other looks good to me!

docs/analytical_plan.md
swo: yes, this should read "in the future relative to the forecast date", or just "after the forecast date"

docs/analytical_plan.md
swo: Yes, in this (and many other) projects we're optimistically assuming, for the purposes of retrospective evaluation, that the report date is equal to the observation date. I'll make that clear.

In many data sets, of which NIS is one, the report date is not explicitly recorded. The NIS weekly data are tagged with the end of a week, and surveys happened all through that week. But it takes some time for the results to be tabulated and reported. If/when we do prospective forecasting, there will be some lag between the end of a week and when we have access to the data for that week.

It's actually even more complicated, because you get data updates through time. We're assuming that report date = obs date and also that there was only one data point per observation date (i.e., the data were never updated after the first report)

docs/analytical_plan.md
swo: In the simplest forecasting problems, the report date is some time (the "reporting delay") after the reference date, and you get final data. In that case, retrospective forecasting and prospective forecasting are the same, except that the forecasts can only come out later (i.e., the reporting delay).

In medium-difficulty problems, the data will update. So then each reference date has multiple report dates. This might happen if there were, say, a data error that gets fixed. When you do a prospective forecast, you use all the most recently available data, i.e., the most recent report date for each reference date. When doing retrospective forecasting, you would use the data that was available as of the forecasting date, i.e., the latest report date before the forecasting date.

Things get harder for counts and nowcasting problems. Say you're interested in the number of people who died of Covid on reference date R. On day R+1, a very small number of those deaths will have been reported. On day R+2, a few more will come in, etc. It's easy to do retrospective forecasting badly, and use the finally-reported number of deaths, rather than the number that were available as of some date. It's also means you need to treat more recent reference dates differently, because you know they will be biased toward incomplete reporting.

Issue #30, opened 2024-10-25
Make rollout date a model parameter, not data preprocessing


eschrom:
The code currently assumes that every data set used for model training will have a known rollout date for every region represented. Will there be cases where rollout date is unknown and ought to be fit? Or cases where rollout date is not a relevant piece of information and should be ignored? If so, these cases must be defined and accounted for.

swo:
For Covid, there truly is a rollout date, when the thing was first available

For flu, there's a long leaky startup

For now, we assert rollout dates as a matter of data preprocessing

In the future, we might want to make rollout an adjustable parameter for some models (or a required, fixed input for other models)

eschrom:
The logistic plus linear model does not need to know or infer a rollout date. The concept of a rollout date is removed from the codebase as of #164. Thus, this issue is closed as not planned.

PR #114, opened 2025-02-01
Document explaining model structures and development plans

eschrom:
There are many directions we could take the refining of existing models and/or introduction of new model structures. I have made a document to summarize where we are now, some of the issues observed, and ideas for model development.

### DIFF
--- a/docs/model_development.md
+++ b/docs/model_development.md
@@ -11,6 +11,8 @@ There are two broad types of models under consideration: autoregressive (AR) and
 
 Several facets of each model are discussed, including misspecification, hyperparameters, forecast uncertainty, and developmental priorities.
 
+Importantly, for the first version of each model, the observed cumulative uptake $c_t$, and the incident uptake $u_t = c_t - c_{t-1}$ that it implies, are considered synonymous with true uptake. In future versions of each model, it will be important to separate latent true uptake from the observed uptake. The former is never observed and the latter is subject to error, which is estimated in the NIS data.
+
 # Autoregressive (AR) Models
 
 Let $u_t$ be the incident uptake at time $t$, where $t$ is measured in weeks since rollout.
@@ -48,9 +50,19 @@ Practically, nonsensical predictions as a result of this misspecification can be
 - Priors for $\alpha$ and $\beta_x$ should have a zero mean and small variance.
 - The prior for $\sigma$ should have a small expectation.
 
-Under these conditions, parameter combinations that predict nonsensical future incident uptake are unlikely to be drawn from the posterior.
+Under these conditions, parameter combinations that predict nonsensical future incident uptake are unlikely to be drawn from the posterior. Nonetheless, refactoring the linear incident uptake model to avoid misspecification altogether is desirable.
 
-Nonetheless, refactoring the linear incident uptake model to avoid misspecification altogether is desirable. Here is an idea...
+The best approach may involve separating latent true uptake (cumulative $\hat{c}_t$ and incident $\hat{u}_t$) from observed uptake (cumulative $c_t$ and incident $u_t$), and making use of the observed variability ($\sigma_t$) that is also reported alongside $c_t$ in the data. This allows the natural limits of $[0,~1]$ to be imposed on the observed uptake $c_t$, via a truncated normal distribution (denoted $TruncNorm(\text{mean, variance, lower bound, upper bound})$):
+
+$$
+\begin{align*}
+&c_t \sim TruncNorm(\hat{c}_t, \sigma_t, 0, 1) \\
+&\hat{c}_t = \sum_{i=0}^{t} \hat{u}_i \\
+&\hat{u}_i \sim N(\mu_i, \sigma) \\
+&\mu_i = \alpha + \beta_{u}\hat{u}_{i-1} + \beta_{t}t + \beta_{tu}t\hat{u}_{t-1} \\
+&\alpha,~\beta_x,~\sigma \sim \text{prior distributions} \\
+&\end{align*}
+$$
 
 An alternative formulation might replace the incident uptake $u_t$ with the fractional incident uptake $f_t$, where $f_t$ is the _fraction_ of the unvaccinated population in week $t-1$ who got vaccinated in week $t$. Mathematically, $f_t = \frac{u_t}{1-c_{t-1}} = \frac{c_t - c_{t-1}}{1-c_{t-1}}$. Notice that $f_t$ is bounded by $[0,~1]$. Thus, the following model would not be misspecified:
 
--- a/docs/model_development.md
+++ b/docs/model_development.md
@@ -54,7 +54,7 @@ Under these conditions, parameter combinations that predict nonsensical future i
 
 The best approach may involve separating latent true uptake (cumulative $\hat{c}_t$ and incident $\hat{u}_t$) from observed uptake (cumulative $c_t$ and incident $u_t$), and making use of the observed variability ($\sigma_t$) that is also reported alongside $c_t$ in the data. This allows the natural limits of $[0,~1]$ to be imposed on the observed uptake $c_t$, via a truncated normal distribution, denoted $TruncNorm(\text{mean, variance, lower bound, upper bound})$.
 
-$$
+```math
 \begin{align*}
 &c_t \sim TruncNorm(\hat{c}_t, \sigma_t, 0, 1) \\
 &\hat{c}_t = \sum_{i=0}^{t} \hat{u}_i \\
@@ -62,7 +62,7 @@ $$
 &\mu_i = \alpha + \beta_{u}\hat{u}_{i-1} + \beta_{t}t + \beta_{tu}t\hat{u}_{t-1} \\
 &\alpha,~\beta_x,~\sigma \sim \text{prior distributions} \\
 \end{align*}
-$$
+```
 
 Even still, $\hat{u}_i$ might escape $[0,~1-\hat{c}_i]$, so $\hat{c}_t$ might escape $[0,~1]$. The truncated normal distribution for $c_t$ prevents this from ultimately manifesting in nonsensical observed cumulative uptake, but some misspecification still lurks.
 
--- a/docs/model_development.md
+++ b/docs/model_development.md
@@ -52,7 +52,7 @@ Practically, nonsensical predictions as a result of this misspecification can be
 
 Under these conditions, parameter combinations that predict nonsensical future incident uptake are unlikely to be drawn from the posterior. Nonetheless, refactoring the linear incident uptake model to avoid misspecification altogether is desirable.
 
-The best approach may involve separating latent true uptake (cumulative $\hat{c}_t$ and incident $\hat{u}_t$) from observed uptake (cumulative $c_t$ and incident $u_t$), and making use of the observed variability ($\sigma_t$) that is also reported alongside $c_t$ in the data. This allows the natural limits of $[0,~1]$ to be imposed on the observed uptake $c_t$, via a truncated normal distribution (denoted $TruncNorm(\text{mean, variance, lower bound, upper bound})$ ):
+The best approach may involve separating latent true uptake (cumulative $\hat{c}_t$ and incident $\hat{u}_t$) from observed uptake (cumulative $c_t$ and incident $u_t$), and making use of the observed variability ($\sigma_t$) that is also reported alongside $c_t$ in the data. This allows the natural limits of $[0,~1]$ to be imposed on the observed uptake $c_t$, via a truncated normal distribution, denoted $TruncNorm(\text{mean, variance, lower bound, upper bound})$.
 
 $$
 \begin{align*}
--- a/docs/model_development.md
+++ b/docs/model_development.md
@@ -22,9 +22,9 @@ A generic AR model of the I<sup>th</sup> order has the form:
 $$
 \begin{align*}
 &u_t \sim N(\mu_t, \sigma) \\
-&\mu_t = \sum_{i=1}^{I} \beta_{i}u_{t-i} \\
+&\mu_t = \sum_{i=0}^{t} \beta_{i}u_{i} \\
 &\beta_{i},~\sigma \sim \text{prior distributions} \\
-&\end{align*}
+\end{align*}
 $$
 
 The "linear incident uptake model" currently implemented is a first-order AR model, but with the added complexity of absolute time as a predictor, too:
@@ -34,7 +34,7 @@ $$
 &u_t \sim N(\mu_t, \sigma) \\
 &\mu_t = \alpha + \beta_{u}u_{t-1} + \beta_{t}t + \beta_{tu}tu_{t-1} \\
 &\alpha,~\beta_x,~\sigma \sim \text{prior distributions} \\
-&\end{align*}
+\end{align*}
 $$
 
 Essentially, this model says that the uptake rate changes through time ($\beta_t$) with some inertia ($\beta_u$), and that the inertia itself can change through time ($\beta_{tu}$).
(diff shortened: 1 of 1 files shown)


docs/model_development.md
```diff
@@ -0,0 +1,149 @@
...
+# Autoregressive (AR) Models
+
+Let $u_t$ be the incident uptake at time $t$, where $t$ is measured in weeks since rollout.
+
+A generic AR model of the I<sup>th</sup> order has the form:
+
+$$
+\begin{align*}
+&u_t \sim N(\mu, \sigma) \\
+&\mu = \sum_{i=1}^{I} \beta_{i}u_{t-i} \\
```
swo, line 23: I think this is $\mu_t$?

More broadly, and perhaps needing to return to your 3 pictures of how AR models work, I would have expected:

```math
\begin{gather*}
c_t \sim N(\hat{c}_t, \sigma_t) \\
\hat{c}_t = \sum_{t'=0}^t \hat{u}_t \\
\hat{u}_t = \sum_{i=0}^\infty \beta_i \hat{u}_{t-i} + \epsilon_t
\end{gather*}
```

where:

- $c_t$ is observed cumulative uptake
- $\sigma_t$ is "observed" data variability (ie confidence interval as reported by NIS)
- $\hat{c}_t$ is the estimate of $c$, which is equivalently the underlying "true" cumulative uptake
- similarly for $\hat{u}$ -- note that I don't talk about plain $u$, because finding the observation error on $u$ is not trivial (not saying we couldn't find *some* way to estimate it, but you can't just read it off from the CI from the data)
- the $\beta_i$ are the autoregressive behavior
- $\epsilon_t$ are non-autoregressive (random walk) pieces
- we need priors for $\hat{u}_0$, $\beta_i$, and whatever parameterizes $\epsilon_t$
- there needs to be some restriction that $\hat{u}_t \geq 0$ for all $t$ -- one way would be to say that the prior on $\hat{u}_0$ must have no weight for negative values (which we should do anyway) and also that $\beta_i > 0$ for all $i$ (but this seems too restrictive)

I think the advantages here are:

- Clarify that observation process is on $c$ while the regressive process is on $u$
- Separate observation error from random walks: if we get an anomalously high data point for some week, we could either put that into observation noise ($u_t$ differing from $\hat{u}_t$) or into some non-autoregressive effect $\epsilon_t$

Because we don't directly constrain $c$, there could be a decrease in cumulative uptake between the last observed date and (the median of forecasts from) the first forecast date. A big discontinuity is almost certainly a bad fit. But we should allow the model to assert that the last cumulative uptake data point was an *observation* anomaly.

docs/model_development.md
```diff
@@ -0,0 +1,149 @@
+# Purpose
+
+This is a summary of the vaccine uptake models under consideration, as well as some issues and development priorities for each.
+
+# Basic Structures
+
+There are two broad types of models under consideration: autoregressive (AR) and full-curve (FC) models.
+
+- AR: Incident uptake at time $t$, $u_t$, is a function previous incident uptake value(s), among other predictors.
+- FC: Cumulative uptake at time $t$, $c_t$, is a function of time, among other predictors.
```
swo, line 10: I think the more typical phrasing is "parametric curve fitting," but that might be jargon I made up

docs/model_development.md
```diff
@@ -0,0 +1,149 @@
...
+&\mu = \alpha + \beta_{u}u_{t-1} + \beta_{t}t + \beta_{tu}tu_{t-1} \\
+&\alpha,~\beta_x,~\sigma \sim \text{prior distributions} \\
+&\end{align*}
+$$
+
+Essentially, this model says that the uptake rate changes through time ($\beta_t$) with some inertia ($\beta_u$), and that the inertia itself can change through time ($\beta_{tu}$).
+
+## Misspecification
+
+Incident uptake $u_t$ is naturally bounded by $[0,~1-c_{t-1}]$, meaning that over the course of a week $t$, the number of people who get vaccinated cannot be fewer than 0 or exceed the number of people who were unvaccinated in week $t-1$
```
swo, line 42: Define $c_t$

docs/model_development.md
```diff
@@ -0,0 +1,149 @@
...
+\begin{align*}
+&f_t \sim Beta(\mu, \nu) \\
+&\text{logit}(\mu) = \alpha + \beta_{f}f_{t-1} + \beta_{t}t + \beta_{tf}tf_{t-1} \\
+&\alpha,~\beta_x,~\nu \sim \text{prior distributions} \\
+&\end{align*}
+$$
+
+The logit link function prevents $\mu$ from escaping the bounds $(0,~1)$, and the Beta likelihood also cannot escape these bounds.
+
+One problem remains. Although _true_ fractional incident uptake $f_t$ is never less than 0, the _true_ uptake is not the same as the _reported_ uptake. In fact, there are instances of negative uptake reported in the data. To handle this, a separate term for observation error is necessary. If Gaussian error is assumed, the problem of misspecification is reintroduced (though perhaps with minimal consequences). Further thought is necessary.
```
swo, line 67: See my note above about observation process on $c$ and regressive process on $u$. I think this resolves some of this trouble.

docs/model_development.md
swo: Agree that this is annoying. But this must be something that other forecasting models face: a hospitalization forecast could predict more hospitalizations than people for some time period.

My sense is that the observation process should account for this. Maybe the pdf on the observed cumulative uptake $c$ as a function of the estimated underlying cumulative uptake $\hat{c}$ should be:

$$
f_{C_t}(c_t) = \begin{cases}
0 & c_t < 0 \\
f_\mathrm{Norm}(c_t; \hat{c}_t, \sigma_t) & 0 \leq c_t \leq 1 \\
0 & c_t > 1
\end{cases}
$$

So the posterior should produce no forecasts that have values outside the correct range. (I'm not sure if normal Bayesian model fitting can produce this, because there is some complicated *combination* of parameters that produces uptakes outside the relevant range, and it's maybe not easy to ablate those parts of the posterior.)

docs/model_development.md
```diff
@@ -0,0 +1,149 @@
...
+Practically, nonsensical predictions as a result of this misspecification can be avoided by choosing priors carefully. In particular,
+
+- Priors for $\alpha$ and $\beta_x$ should have a zero mean and small variance.
+- The prior for $\sigma$ should have a small expectation.
+
+Under these conditions, parameter combinations that predict nonsensical future incident uptake are unlikely to be drawn from the posterior.
+
+Nonetheless, refactoring the linear incident uptake model to avoid misspecification altogether is desirable. Here is an idea...
+
+An alternative formulation might replace the incident uptake $u_t$ with the fractional incident uptake $f_t$, where $f_t$ is the _fraction_ of the unvaccinated population in week $t-1$ who got vaccinated in week $t$. Mathematically, $f_t = \frac{u_t}{1-c_{t-1}} = \frac{c_t - c_{t-1}}{1-c_{t-1}}$. Notice that $f_t$ is bounded by $[0,~1]$. Thus, the following model would not be misspecified:
```
swo, line 55: I feel the same temptation, but I worry this will be more pain than it's worth. What does it mean, mathematically or epidemiologically, to have a model that predicts that 10% of unvaccinated people will get vaccinated in a given week?

On the other hand, if these kinds of models fit easily, and ultimately don't look that much different from the fits we get from other places, maybe it's feasible to reinterpret these parameters in some more sensible way. So maybe something to put on the backlog, as another form of model.

swo (commented):
I'm only half-way through, but wanted to get some thoughts started!

Fuhan-Yang (approved):
Thanks Ed for coming up with model ideas! My comments are research questions, which should not gate the model development.

docs/model_development.md
```diff
@@ -0,0 +1,149 @@
...
+$$
+
+Here, $\alpha$ and $\beta_x$ are hyperparameters: they govern hyperdistributions, from which a separate draw is made for each season.
+
+Factors other than season could also be used to group the data. For example, further hyperparameters could be used to infer different parameter values (drawn from a common distribution) that describe the uptake in different geographic regions or among different demographic groups.
+
+## Forecast Uncertainty
+
+Forecasting with AR models naturally produces a cone of uncertainty that expands into the future. Each draw from the posterior distribution is a unique combination of parameter values that defines a trajectory of uptake going forward (still with some stochastic influence on observations, from $\sigma$). All these trajectories sprout from the last observed data point and diverge as they move into the future.
+
```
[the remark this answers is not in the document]
swo: See my note above about distinguishing the observed uptake data from the latent uptakes. The cone of uncertainty expands because of the combinations of uncertainties on $\hat{u}_0$, $\beta_i$, and $\epsilon_t$

docs/model_development.md
```diff
@@ -0,0 +1,149 @@
...
+&A_s \sim N(A,~\sigma_A) \\
+&H_s \sim N(H,~\sigma_H) \\
+&n_s \sim N(n,~\sigma_n) \\
+&A,~H,~n,~\sigma_A,~\sigma_H,~\sigma_n,~\sigma \sim \text{prior distributions} \\
+\end{align*}
+$$
+
+And again, factors other than season, such as geographic area or demographic group, could be used to group the data.
+
+## Forecasting Uncertainty
```
swo, line 136: I think parametric curve fitting should still work... you get the greatest uncertainty on $\hat{u}$ in places where you don't have data, ie in the future. It's like any regression, that error bars are largest as you project further away from the data

docs/model_development.md
eschrom: Well I definitely made up the "full curve" jargon haha. As long as it is clear what we are talking about, and how it differs from AR, the specific jargon doesn't matter to me.

docs/model_development.md
eschrom: Again, this document does not (yet) consider that observed cumulative uptake differs from true cumulative uptake, because we have yet to make this distinction in a model. So $c_t$ is defined in line 10. But I agree that we should move toward this distinction, likely following the framework you laid out in your comment on Line 23. 

That framework partially solves misspecification, and this adjustment of using a truncated Normal distribution for $c_t$ would finish the job. But truncated Normals, especially where the truncation depends on $t$, may be more trouble than they are worth. I imagine our forecasting will almost always be in the regime where cumulative uptake has already been observed to have climbed above 0, and it will never get close to 1. So the truncation may offer very little benefit in practice.

docs/model_development.md
eschrom: I agree!

docs/model_development.md
```diff
@@ -0,0 +1,149 @@
...
+
+## Hyperparameters
+
+Realistically, the $\alpha$ and $\beta_x$ parameters that control uptake rate probably differ slightly but systematically from season to season (where "season" refers to an overwinter disease season, e.g. "2024/2025"). And multiple past seasons of data are available for training models (two past seasons for covid, and many more for flu). Therefore, it would be wise to introduce hyperparameters with a subscript "s" for season:
+
+$$
+\begin{align*}
+&u_t \sim N(\mu, \sigma) \\
+&\mu = \alpha_s + \beta_{u,s}u_{t-1} + \beta_{t,s}t + \beta_{tu,s}tu_{t-1} \\
+&\alpha_s \sim N(\alpha,~\sigma_{\alpha}) \\
```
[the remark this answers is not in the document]
swo: This is the same as #30 , I think

docs/model_development.md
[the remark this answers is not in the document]
eschrom: Even before we distinguish true latent uptake from observed uptake, it is worth clarifying how uncertainty compounds into the future when forecasting with our current AR model. From each $[\alpha, \beta_x, \sigma]$ sample from the posterior distribution $[\alpha, \beta_x]$ defines a deterministic trajectory into the future. But at each timepoint along the way, there is still uncertainty thanks to $\sigma$. So at each timepoint, one value is sampled, to serve as the basis of projecting the next timepoint, and so on. 

This diagram may help:
image

image

docs/model_development.md
eschrom: Yep - I plan to clarify that before merging!

docs/model_development.md
```diff
@@ -0,0 +1,149 @@
...
+\begin{align*}
+&c_t \sim N(\mu, \sigma) \\
+&\mu = \frac{A_s~t^{n_s}}{H_s^{n_s} + t^{n_s}} \\
+&A_s \sim N(A,~\sigma_A) \\
+&H_s \sim N(H,~\sigma_H) \\
+&n_s \sim N(n,~\sigma_n) \\
+&A,~H,~n,~\sigma_A,~\sigma_H,~\sigma_n,~\sigma \sim \text{prior distributions} \\
+\end{align*}
+$$
+
```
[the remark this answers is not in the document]
eschrom: Good point! In principle, you are correct. In practice, I imagine that a thin enough prior distribution for $A$ and $\sigma_A$ would prevent predicted values of $c_t$ from escaping their bounds.

docs/model_development.md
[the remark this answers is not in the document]
eschrom: Yep! That is #30.

docs/model_development.md
eschrom: I disagree. Think of the Hill model parameters: A = maximum uptake, H = half-maximal time, and n = steepness. Uncertainty at the end of the season depends mostly on uncertainty in A. Uncertainty at the middle of the season depends on uncertainty in A, H, and n.

Moreover, if we are fitting on 15 past seasons of data and the first half of this season, that means we have 16 data points per timepoint in early and midseason, and 15 data points per timepoint at the end of season. That's nearly the same amount of data to fit with, no matter where you are along the curve.

This is an old figure, but...   The red curve here is a Hill model, fit to the flu data before ~Oct 2024, and then projecting ~Nov 2024 onward. (There are no hyperparameters for season here.) Note that the 95% credible interval starts somewhat wide, even right where the observations leave off, and it does not expand much into the future.
image

docs/model_development.md
swo: It seems not hard to put truncation into a Numpyro model: https://num.pyro.ai/en/0.15.3/tutorials/truncated_distributions.html#4.-Ready-to-use-truncated-and-folded-distributions

I guess I find this much easier than moving into logit space or something

eschrom: Fair, and this is nice because $c_t$ is always bounded by 0 and 1, as opposed to $u_t$ for which the upper bound changes as the uptake process moves along.

docs/model_development.md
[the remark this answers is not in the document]
eschrom: On closer inspection, your comments are about #30 here, but this section is not. This section is about allowing the model parameters to differ by season, to account for different seasons having slightly different curve shapes.

docs/model_development.md
eschrom: Remember, there is no distinction between latent true uptake and noisy observed uptake in these models, yet. 

I do not see this as desirable behavior. Even if the Hill model matched the data perfectly at the junction between data and forecast (i.e. at the forecast date), the wide uncertainty there says that cumulative uptake could plausibly go down ~5% in the week after the forecast date. That's clearly not right.

I believe the root of the problem is this. The autoregressive model asks, "given what I know about this week, what can I say about next?" Uncertainty therefore compounds as you move into the future. On the other hand, the Hill model asks, "what is the _full curve_ likely to look like?" There may be just as much uncertainty about the middle of the full curve as there is about the tails.

I am soon to submit another PR which I hope will help illustrate this a bit better.

eschrom:
I think this doc has already done its job, which is to stimulate thoughtful discussion about the next steps in model development. In particular, I think the autoregressive model that @swo outlined, separating latent true uptake from noisy observed uptake, should be the next step.

So I plan to merge this doc and to edit it periodically in the future, whenever we need a catalyst for more model development brainstorming.

PR #115, opened 2025-02-03
Plot projections and eval scores

Fuhan-Yang:
Add diagnostic plots of time-varying projections compared with data, and evaluation scores differed by forecast dates. Minor bugs showed up when using different forecast periods. Worth to try multiple forecast periods to ensure the robustness of the pipeline.


scripts/proj_plot.py
```diff
@@ -0,0 +1,93 @@
...
+
+    obs_chart = (
+        alt.Chart(obs)
+        .mark_circle(color="black")
+        .encode(alt.X("time_end", axis=time_axis), alt.Y("estimate"))
+    )
+
+    charts = []
+
+    forecast_starts = [date for date in pred["forecast_start"].unique()]
```
eschrom, line 43: Should we check that the `pred` frame is already sorted by `forecast_start` to make sure that the plot panels come out in a sensible order? I am unfamiliar with altair plotting, so maybe panel ordering is done automatically - just wanted to check.

scripts/proj_plot.py
```diff
@@ -0,0 +1,93 @@
...
+    """
+    Save a multiple-grid graph with the comparison between the observed uptake and the prediction,
+    initiated over the season.
+
+    Arguments:
+    --------------
+    obs: polars.Dataframe
+        The observed uptake data frame, indicating the cumulative vaccine uptake as of `time_end`.
+    pred: polars.Dataframe
+        The predicted daily uptake, differed by forecast date, must include columns `forecast_start` and `estimate`.
```
eschrom, line 18: I would say, "The predicted cumulative uptake, ..." since the predictions taken here are cumulative, whereas daily average uptake is a separate that actually appears in some of the model fitting.

scripts/proj_plot.py
```diff
@@ -0,0 +1,93 @@
...
+    config: dict
+        config.yaml to define the number of columns in the graph, the saving path and the saving name.
```
eschrom, line 20: The config doesn't actually have the path or name for saving, does it?

scripts/score_plot.py
```diff
@@ -0,0 +1,60 @@
...
+def plot_score(scores):
+    """
+    Save a evaluation score plot, varied by forecast start date.
+
+    Arguments:
+    --------------
+    scores: polars.DataFrame
+        The evaluation scores data frame.
+    config: dict
+        config.yaml to define the saving path and the saving name.
```
eschrom, line 17: Again, the config doesn't really have the path or name for saving, right? I thought those were in the makefile. The config has the full names of the evaluation metrics, though.

eschrom (changes requested):
Awesome work, @Fuhan-Yang! I apologize that I'm not well-versed in altair, so I cannot review the actual plotting with much detail. But I did find one spot that I think needs a change - the forecasts are saved in long, not wide, format now. Otherwise, there are just some questions peppered throughout.

scripts/proj_plot.py
```diff
@@ -0,0 +1,93 @@
...
+    args = p.parse_args()
+
+    with open(args.config, "r") as f:
+        config = yaml.safe_load(f)
+
+    pred = pl.scan_parquet(args.pred).collect()
+    data = pl.scan_parquet(args.obs).collect()
+
+    # Drop all samples and just use mean estimate, for now
+    pred = pred.drop([col for col in pred.columns if "estimate_" in col])
```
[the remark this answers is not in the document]
swo: I forget if this was being discussed elsewhere, but I have a strong preference for long format for these things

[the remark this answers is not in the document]
eschrom: It was, and we fixed it!

scripts/proj_plot.py
swo: altair won't mind if the rows in the input data frame aren't sorted in the order that the points will be left-to-right plotted

swo (changes requested):
My big things are:

- Making sure that config is *not* passed to the postprocessing plot scripts
- Simplifying the plots using existing facet or grid systems

I also don't think it's crazy to have a single `postprocessing.py` that plots the forecasts and the scores. Because at some point we'll also want some csv (or whatever) reports of some of these values. Rather than have a different script for each of them, we could just stick them all together. There aren't *that* many outputs being made, that this shouldn't be too slow? (Unless it *is* slow to make some of these plots, in which case, we should keep them separate!)

scripts/proj_plot.py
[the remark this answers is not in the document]
Fuhan-Yang: Thanks for catching that! This line may adapt from previous version and didn't raise error because the condition didn't meet. We indeed switched to long format, and just average over for now. This line is deleted.

scripts/proj_plot.py
Fuhan-Yang: Fixed!

scripts/proj_plot.py
Fuhan-Yang: No, this was my trial but it got deprecated. Thanks for catching that

scripts/score_plot.py
Fuhan-Yang: Fixed. Thanks for catching

scripts/config_template.yaml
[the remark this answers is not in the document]
Fuhan-Yang: > But "plot: True" is something that should really be in the Makefile, and how many columns in the subplot is something we can just leave in the script and tweak there as needed

Thanks for pointing out what are important for config file, which I found vague so far. How to put "plot: True" in Makefile to let users choose to plot or not? I will put the number of subplot columns in the script, and would like you to demo the plotting option in Makefile!

docs/model_development.md
```diff
@@ -0,0 +1,161 @@
+# Purpose
```
eschrom, line 1: Why does this doc show up as a new file as of this PR, when it has already been merged as part of #114? Just make sure to rebase before merging, in case this is an older version of this doc - wouldn't want to overwrite the merged version with an older version.

scripts/postprocess.py
```diff
@@ -0,0 +1,139 @@
...
+    p.add_argument("--score_output", help="png file of score plot")
+    args = p.parse_args()
+
+    with open(args.config, "r") as f:
+        config = yaml.safe_load(f)
+
+    pred = pl.scan_parquet(args.pred).collect()
+    data = pl.scan_parquet(args.obs).collect()
+
+    if config["projection_plot"]["plot"]:
```
eschrom, line 133: I don't see a section of the config labeled "projection_plot" or "score_plot," but I also don't see any updates to the config in this PR. Is an updated config missing from this PR? 

That said, @swo had indicated that config entries to control which plots are generated was not a desirable strategy. How did that conversation resolve?

Maybe a good strategy is just to have this script do everything if `config["evaluation_timeframe"]["interval"] is not None` (i.e. if evaluation is being performed) and do nothing otherwise, since this is all focused on retrospective forecasting anyway.

scripts/postprocess.py
```diff
@@ -0,0 +1,139 @@
...
+    obs_chart = (
+        alt.Chart(obs)
+        .mark_circle(color="black")
+        .encode(alt.X("time_end", axis=time_axis), alt.Y("estimate"))
+    )
+
+    pred = pred.with_columns(
+        date_str=("Forecast Date:" + pl.col("forecast_start").cast(pl.Utf8))
+    )
+    obs_chart = (
```
eschrom, line 44: Looks like you are overwriting what was done in lines 35-39?

scripts/postprocess.py
```diff
@@ -0,0 +1,139 @@
...
+        alt.Chart(obs)
+        .mark_point()
+        .encode(
+            x="time_end:T",
+            y="estimate:Q",
+        )
+    )
+
+    pred_chart = (
+        alt.Chart()
```
eschrom, line 54: Doesn't `pred` have to be given to `alt.Chart()`?

eschrom (changes requested):
I still have a few questions before this gets merged. It's getting close, though! I like that it's a single script for all evaluation plots now.

scripts/postprocess.py
Fuhan-Yang: Good catch! I removed the plotting option in config as suggested by @swo, but still keep the original in the plotting script. This can be good option, but that conversation hasn't been resolved, I'm curious how to change the Makefile to have the plotting option!

scripts/postprocess.py
Fuhan-Yang: Thanks! Just updated

scripts/postprocess.py
Fuhan-Yang: No, it seems like the way to have faceted plots in altair is to pass the data for faceting at `alt.layer()` later

Fuhan-Yang:
Just put the plotting option back in config, ensuring the current version is runnable

scripts/postprocess.py
eschrom: Okay - good to know!

scripts/postprocess.py
eschrom: Sounds good to me. I suspect we'll eventually pull these arguments out of the config, but I don't think that has to gate us here. I am eager to run and debug the full pipeline. As long as this allows everything to run including plotting, I'm happy to address it later.

eschrom (approved):
As noted, I think there are still a few improvements to be made, but that goes for the whole pipeline and not just this PR. I am happy for you to merge. Then I'll start scouring for bugs!

swo (changes requested):
Kindly rebase so we can double-check exactly what's in here

And then I made suggestions in #119 that I think should get incorporated here, or discussed as the next PR

Fuhan-Yang:
It is rebased! Only commits related to this PR are included

scripts/postprocess.py
swo: My point is to just always have the plots as part of the workflow `make all`, so you don't need a yes/no choice anywhere

you could `make data/scores.parquet`, if you wanted to run everything up through the scores but nothing after

`make output/projections.png` should run a command that actually causes `output/projections.png` to get made. So it's weird to call a script and give it an option to do nothing.

swo:
After merging in #119 , I have most of the things I was hoping for!

Issue #91, opened 2025-01-02
Align cumulative Hill model with latest data


eschrom:
A Hill model fit directly on cumulative uptake does not account for seasonal variation - it only finds the "average" uptake S curve. So for any specific season, in-season projections do not pick up right where the data left off. This must be corrected. Below is an idea from @swo: 

> Or, maybe, the model should be looking at cumulative uptake after the last observed data point? Like, if you observe cumulative u t up > to timepoint t ⋆ , and the cumulative uptake is a function f ( t ) , then your projections are u ( t ⋆ ) + [ f ( t ) − f ( t ⋆ ) ] or something

eschrom:
This is largely but not completely addressed by incorporating season in #117

PR #117, opened 2025-02-05
Hyperparameters for season in the Hill model (in R demo only)

eschrom:
I have updated the R demo to incorporate hyperparameters into the Hill model that allow the final uptake and the half-maximal date to vary by season. Because this is just in the R demo, which is for brainstorming and will not be part of the final codebase, it doesn't require a detailed review. This is just to demonstrate few points, using the flu data.

As a reminder, here is what the entirety of the available flu data looks like:
image

If a Hill model is fit to these data, and then a forecast for the 2024/25 season is made from late October 2024 onward, we get:
image
Notice how 1) the forecast does not align with the data on the forecast date, and 2) the 95% credible interval starts off wide without widening much into the future.

Now we do the same thing, but including hyperparameters to allow the exact shape of the Hill curve to vary by season:
image
Both problems are nearly solved! The forecast aligns closely (though not perfectly) with the data on the forecast date, and the credible interval starts off much thinner (though still encompassing some values lower than observed cumulative uptake on the forecast date).

We can probably ignore the last vestiges of these two problems, because we are moving toward a refactor in which we no longer trust the observed data to be synonymous with the true latent uptake. In that case, some uncertainty about what the true uptake is, even on dates that have already been observed, is totally acceptable.

I think this is a nice validation of @swo's intuition for how prediction uncertainty should be greater at the tail of a sigmoid curve as opposed to the middle, and a nice validation of our collective intuition that grouping the data on season is important.

### DIFF
--- a/scripts/model_demo.r
+++ b/scripts/model_demo.r
@@ -21,7 +21,7 @@ library(lubridate)
 library(brms)
 
 # Set the user-controlled parameters
-disease <- "covid" # "covid" or "flu"
+disease <- "flu" # "covid" or "flu"
 forecast_date <- as.Date("2024-10-25")
 end_date <- as.Date("2025-05-31")
 
--- a/scripts/model_demo.r
+++ b/scripts/model_demo.r
@@ -140,9 +140,15 @@ train_lim$elapsed_std <- scale(train_lim$elapsed)[, 1]
 train_lim$previous_std <- scale(train_lim$previous)[, 1]
 
 # Build the LIM
+priors <- c(
+    prior(normal(0, 0.2), class = "Intercept"),
+    prior(normal(0, 0.2), class = "b"),
+    prior(cauchy(0, 0.5), class = "sigma")
+)
 lim <- brms::brm(daily_std ~ previous_std * elapsed_std,
     data = train_lim,
-    family = gaussian()
+    family = gaussian(),
+    prior = priors
 )
 
 # Set up data frame for LIM projection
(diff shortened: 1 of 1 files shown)


eschrom:
This largely addresses #91 by the way.

swo:
I guess what's going on here is that, if the prior on the hyperparameters are poorly specified, then the model looks at this year of data and can conclude that every single data point so far this season is an error, and the *true* underlying uptake is actually closer to what we've seen in previous years

Remind me how this model deals with observation error? If the observation error is too "loose," then I imagine the model is more prone to not matching the observed data from this season

eschrom:
Yes! That first paragraph is an _excellent_ way to describe what's "wrong" with this model. Second paragraph too. Super clear! 

This model does not really deal with observation very explicitly. The exact formulation is: 

$$
\begin{align*}
&c_t \sim N(\mu_t, \sigma) \\
&\mu_t = \frac{A_s~t^{n}}{H_s^{n} + t^{n}} \\
&A_s \sim N(A,~\sigma_A) \\
&H_s \sim N(H,~\sigma_H) \\
&A,~H,~n,~\sigma_A,~\sigma_H,~\sigma \sim \text{prior distributions} \\
\end{align*}
$$

As with the autoregressive model, I built the v.1 Hill model super simple to help demonstrate some key points (e.g. the importance of parameter variation by season, here). And as with the autoregressive model, we should refactor in the future to deal with observation error.

swo:
This all seems sensible, if I'm reading this right: $c_t$ is the reported value, and $\mu_t$ is the latent/true uptake, then $\sigma$ describes the observation error.

My own tweak would be to draw $\sigma$ from the data (i.e., NIS's own estimations of its error), so that $c_t$ and $\sigma_t$ are data.

I'd be curious if the posterior distribution on $\sigma$ matches up at all with the NIS-estimated sigmas.

swo (approved):
As a demo, this seems fine

Do we have the model description in the docs somewhere? Out of scope for this PR, but I know we'll want it somewhere

eschrom:
Yep, we have a description of this in docs/model_development.md. Indeed, this is merely a demo. I'll make note of your suggested tweak to draw $\sigma$ from the data, because I anticipate building this model as working part of the python code soon.

PR #119, opened 2025-02-12
Suggestions: postprocessing

swo:
- Makefile
  - Do not restart from preprocessing every time (replace `cache` with `$(NIS_CACHE)`)
  - Put in config and ensure it's a proper dependency
  - Remove commented lines
  - Break up long line
- Update altair version: I was getting errors about a mismatch between the versions of Python, `typing_extensions`, and altair that was fixed (for me) by updating altair
- Refactor the plotting script to use built in facetting functionality, rather than building up charts manually
- Remove default config path from scripts; ensure they are put in via the Makefile


Makefile
```diff
@@ -13,21 +13,17 @@ SCORE_PLOTS = output/scores.png
...
+		--proj_output=$(PROJ_PLOTS) --score_output=$(SCORE_PLOTS)
 
 $(SCORES): scripts/eval.py $(FORECASTS)
 	python $< --pred=$(FORECASTS) --obs=$(RAW_DATA) --output=$@
 
 $(FORECASTS): scripts/forecast.py $(RAW_DATA)
 	python $< --input=$(RAW_DATA) --output=$@
 
-$(RAW_DATA): scripts/preprocess.py cache
+$(RAW_DATA): scripts/preprocess.py $(NIS_CACHE)
```
Fuhan-Yang: Does this mean the command line will use the clean data in $(NIS_CACHE) if it is claimed in this way? Otherwise, if it is `cache`, it will redo the preprocess?

Makefile
```diff
@@ -5,19 +5,26 @@ CONFIG = scripts/config.yaml
...
 
-all: $(SCORES)
+all: $(PROJ_PLOTS) $(SCORE_PLOTS)
 
-$(SCORES): scripts/eval.py $(FORECASTS)
-	python $< --pred=$(FORECASTS) --obs=$(RAW_DATA) --output=$@
+$(PROJ_PLOTS) $(SCORE_PLOTS): scripts/postprocess.py $(FORECASTS) $(RAW_DATA) $(SCORES)
+	python $< \
+		--pred=$(FORECASTS) --obs=$(RAW_DATA) --score=$(SCORES) \
+		--proj_output=$(PROJ_PLOTS) --score_output=$(SCORE_PLOTS)
```
Fuhan-Yang, line 19: It seems like there is no option to generate the plot or not in makefile. Or the option is implicitly showed here, like if we don't call `	python $< \
		--pred=$(FORECASTS) --obs=$(RAW_DATA) --score=$(SCORES) \
		--proj_output=$(PROJ_PLOTS) --score_output=$(SCORE_PLOTS)`, the plot won't be generated. `make all` still generate everything

Makefile
```diff
@@ -17,14 +17,14 @@ $(PROJ_PLOTS) $(SCORE_PLOTS): scripts/postprocess.py $(FORECASTS) $(RAW_DATA) $(
 		--pred=$(FORECASTS) --obs=$(RAW_DATA) --score=$(SCORES) \
 		--proj_output=$(PROJ_PLOTS) --score_output=$(SCORE_PLOTS)
 
-$(SCORES): scripts/eval.py $(FORECASTS)
-	python $< --pred=$(FORECASTS) --obs=$(RAW_DATA) --output=$@
+$(SCORES): scripts/eval.py $(FORECASTS) $(CONFIG)
```
Fuhan-Yang, line 20: Why do we remove the default config in the script, and put it in Makefile?

Fuhan-Yang (approved):
Thanks for fixing the PR, my comments are questions, which shouldn't gate merging. I successfully rebased `fy_plot`, so this PR can be rebased

Makefile
swo: `cache` was just a typo. Make was looking for a file named "`cache`", and not finding it, so it figured that this step had failed, which is why `make all` would run this step (and everything after it) every time

Makefile
swo: Correct, `make all` should make all things.

If you wanted to make only some things, you can `make [WHATEVER FILE]` to make just that files (and its dependencies that need to get rebuilt).

Basically `make` gets confused if there is a rule `[target]: [dependencies]` where the rule command doesn't actually make the target

Makefile
swo: there should only be one source of truth for where these options are

so it's better to have the scripts be ignorant of the file structure, and have the file names and defaults only in the Makefile

PR #121, opened 2025-02-13
Assert that dates are sorted chronologically

### LINKED ISSUE #43: Check that UptakeData rows are chronologically ordered
Assert statements in the UptakeData "date_to_..." functions to make sure that the date column is chronologically sorted before performing the function.

eschrom:
Super small PR here - just a few lines! Dates (the `time_end` column) is already sorted chronologically before any operations (e.g. intervals between dates, days since rollout, etc.) are applied to it. I've just added a few assert statements to check this.

This will close #43


Fuhan-Yang (approved):
This looks good to me, thanks Ed!

Issue #88, opened 2024-12-21
Error in running package


Fuhan-Yang:
When running the code under current main, there is an error indicating that there is NaN value in training data. Validation on missing values is needed.

swo:
Do paste in here the error output!

eschrom:
Make sure you are setting the 2023 rollout date to 2023-09-22, rather than 2023-09-01. Covid booster rollout was delayed by a few weeks in 2023. Without this adjustment, the gap between rollout and the first report of that season (2023-09-30) is too long, such that the first report gets dropped from the data. This leaves zero data points in the 2023/2024 season before the first forecast date (2023-10-01), which causes the error you see.

I can confirm that, after the changes in #123, the forecast period of [2023-10-01, 2024-04-30] works. I just ran it with 56 days between each forecast date for speed's sake.

![Image](https://github.com/user-attachments/assets/58c2f36b-26bf-48c6-9bf9-7a533fb27b18)

eschrom:
Closed with #123

PR #123, opened 2025-02-18
Clean up pipeline bugs

eschrom:
There are some bugs in pipeline, which I am searching out and solving. So far, I have found and fixed the following:
- Different filtering terms are required for the historical vs. recent forms of the NIS data within a vaccine type. E.g. for covid, the historical data requires `domain_type` of `overall` and `domain` of `overall`, whereas the recent data requires `domain_type` of `age` and `domain` of `18_ years`. You can now use multiple filtering terms to make sure neither the historical nor the recent data are accidentally filtered out during preprocessing.
- Many operations to prepare the data for model fitting (e.g. converting cumulative to incident) must be performed over some grouping factors. Season is now one of these grouping factors (e.g. so that the incident uptake on the first date of a new season isn't -40%).

### DIFF
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -113,13 +113,13 @@ class UptakeData(Data):
 
 class IncidentUptakeData(UptakeData):
     def to_cumulative(
-        self, group_cols: List[str,], last_cumulative=None
+        self, group_cols: List[str,] | None, last_cumulative=None
     ) -> "CumulativeUptakeData":
         """
         Convert incident to cumulative uptake data.
 
         Parameters
-        group_cols: (str,)
+        group_cols: (str,) | None
             name(s) of the columns of grouping factors
         last_cumulative: pl.DataFrame
             additional cumulative uptake absent from the incident data, for each group
@@ -133,6 +133,9 @@ class IncidentUptakeData(UptakeData):
         such that the sum of incident does not reflect the entire cumulative uptake.
         To fix this, cumulative uptake from the removed rows may be supplied separately.
         """
+        if group_cols is None:
+            group_cols = []
+
         out = self.with_columns(estimate=pl.col("estimate").cum_sum().over(group_cols))
 
         if last_cumulative is not None:
@@ -159,12 +162,12 @@ class CumulativeUptakeData(UptakeData):
             "cumulative uptake `estimate` must be a proportion"
         )
 
-    def to_incident(self, group_cols: List[str,]) -> IncidentUptakeData:
+    def to_incident(self, group_cols: List[str,] | None) -> IncidentUptakeData:
         """
         Convert cumulative to incident uptake data.
 
         Parameters
-        group_cols: (str,)
+        group_cols: (str,) | None
             name(s) of the columns of grouping factors
 
         Returns
@@ -174,6 +177,9 @@ class CumulativeUptakeData(UptakeData):
         Details
         Because the first date for each group is rollout, incident uptake is 0.
         """
+        if group_cols is None:
+            group_cols = []
+
         out = self.with_columns(
             estimate=pl.col("estimate").diff().over(group_cols).fill_null(0)
         )
@@ -181,7 +187,11 @@ class CumulativeUptakeData(UptakeData):
         return IncidentUptakeData(out)
 
     def insert_rollouts(
-        self, rollouts: List[dt.date], group_cols: List[str]
+        self,
+        rollouts: List[dt.date],
+        group_cols: List[str] | None,
+        season_start_month: int,
+        season_start_day: int,
     ) -> "CumulativeUptakeData":
         """
         Insert into cumulative uptake data rows with 0 uptake on rollout dates.
@@ -189,7 +199,7 @@ class CumulativeUptakeData(UptakeData):
         Parameters
         rollout: List[dt.date]
             list of rollout dates
-        group_cols: tuple[str]
+        group_cols: tuple[str] | None
             names of grouping factor columns
 
         Returns
@@ -198,19 +208,31 @@ class CumulativeUptakeData(UptakeData):
         Details
         A separate rollout row is added for every grouping factor combination.
         """
+        if group_cols is None:
+            group_cols = []
+
         frame = self
 
-        if len(group_cols) > 0:
-            rollout_rows = (
-                frame.select(group_cols)
-                .unique()
-                .join(pl.DataFrame({"time_end": rollouts}), how="cross")
-                .with_columns(estimate=0.0)
+        # do not use season as a grouping column to insert rollouts
+        if len([g for g in group_cols if g != "season"]) > 0:
+            rollout_rows = frame.select(group_cols)
+            if "season" in rollout_rows.columns:
+                rollout_rows = rollout_rows.drop("season")
+            rollout_rows = rollout_rows.unique().join(
+                pl.DataFrame({"time_end": rollouts}), how="cross"
             )
-            group_cols = group_cols + ["season"]
         else:
             rollout_rows = pl.DataFrame({"time_end": rollouts, "estimate": 0.0})
-            group_cols = ["season"]
+
+        # add season as a column only after making rollout rows
+        rollout_rows = rollout_rows.with_columns(
+            estimate=0.0,
+            season=pl.col("time_end").pipe(
+                UptakeData.date_to_season,
+                season_start_month=season_start_month,
+                season_start_day=season_start_day,
+            ),
+        )
 
         frame = frame.vstack(rollout_rows.select(frame.columns)).sort("time_end")
 
@@ -221,8 +243,8 @@ class CumulativeUptakeData(UptakeData):
                 frame.with_columns(
                     season=pl.col("time_end").pipe(
                         UptakeData.date_to_season,
-                        season_start_month=min([d.month for d in rollouts]),
-                        season_start_day=min([d.day for d in rollouts]),
+                        season_start_month=season_start_month,
+                        season_start_day=season_start_day,
                     )
                 )
                 .with_columns(
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -113,13 +113,13 @@ class UptakeData(Data):
 
 class IncidentUptakeData(UptakeData):
     def to_cumulative(
-        self, group_cols: List[str,] | None, last_cumulative=None
+        self, group_cols: List[str,], last_cumulative=None
     ) -> "CumulativeUptakeData":
         """
         Convert incident to cumulative uptake data.
 
         Parameters
-        group_cols: (str,) | None
+        group_cols: (str,)
             name(s) of the columns of grouping factors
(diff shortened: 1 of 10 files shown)


eschrom:
More bugs that I have found so far:
- "None" should never be an option for grouping factors. Even if the user does not specify any grouping factors (e.g. geographic region, demographic group, etc.), season always must be a grouping factor. 
- The postprocessing plots should only plot within the season for which forecasting (whether prospective or retrospective) is being done.

eschrom:
And one more bug:
- Projections are made for exactly the dates specified in the config, so if these dates don't match the dates in the test data, evaluation fails. If there are any test data, they are now provided to the projection function, so that forecasting is done for the dates in the test data closest to those specified in the config. Otherwise, forecasting is done for exactly the dates specified in the config, as before.

eschrom:
Because I found several bug (described above), some of which interact with each other, there are a lot of changes here. If you find any of them unclear, please ask questions and I'll be happy to explain!

iup/__init__.py
```diff
@@ -219,7 +219,11 @@ def insert_rollouts(
         assert (
             (
                 frame.with_columns(
-                    season=pl.col("time_end").pipe(UptakeData.date_to_season)
+                    season=pl.col("time_end").pipe(
+                        UptakeData.date_to_season,
+                        season_start_month=min([d.month for d in rollouts]),
+                        season_start_day=min([d.day for d in rollouts]),
+                    )
```
swo, line 226: There's something funny about the logic here: this could end up picking a season start month/day combo that isn't in any of the rollouts

I think the better kludge would be to make season start a config parameter or something? Or refactor `date_to_season` to take a series of season-divider dates, although that's a bigger rework

iup/models.py
```diff
@@ -121,16 +121,17 @@ def extract_starting_conditions(
...
-                .alias("last_cumulative"),
-            ]
+        start = (
+            data.group_by(group_cols)
+            .agg(
+                [
+                    pl.col("time_end").last().alias("last_date"),
+                    pl.col("daily").last().alias("last_daily"),
+                    pl.col("elapsed").last().alias("last_elapsed"),
+                    (pl.col("estimate")).sum().alias("last_cumulative"),
```
swo, line 131:
```suggestion
                    pl.col("estimate").sum().alias("last_cumulative"),
```

iup/models.py
```diff
@@ -328,26 +332,39 @@ def build_scaffold(
...
-                end=end_date,
-                interval=interval,
-                eager=True,
+        # If there is test data such that evaluation will be performed,
+        # use exactly the dates that are in the test data
+        if test_data is not None:
+            scaffold = (
+                test_data.filter(
+                    (pl.col("time_end") >= start_date)
+                    & (pl.col("time_end") <= end_date)
```
swo, line 341: I think you can use `pl.between()` or similar for this

scripts/eval.py
```diff
@@ -27,7 +27,7 @@ def eval_all_forecasts(data, pred, config):
                             pl.col("forecast_start") == forecast_start,
                         )
                     )
-                    .to_incident(config["data"]["groups"])
+                    .to_incident(config["data"]["groups"] + ["season"])
```
swo, line 30: Why not just put "season" into the config?

swo (commented):
I just took a quick skim, so I don't think you should rely me on a check for correctness. Just some clarifying questions about design decisions.

If you and @Fuhan-Yang think this makes sense, I'd say merge it in and proceed forward! There's definitely some refactoring we're going to want to do, but that will be more productive after we have some functioning forecasts

eschrom:
Thanks, @swo! I'll happily incorporate your suggestions, as well as any from @Fuhan-Yang.

scripts/config_template.yaml
```diff
@@ -3,16 +3,16 @@ data:
...
-    domain: 18+ years
-    indicator_type: 4-level vaccination and intent
-    indicator: received a vaccination
-    time_type: week
+    vaccine: [covid]
+    geography_type: [nation]
+    domain_type: [overall, age]
+    domain: [overall, 18+ years]
+    indicator_type: [4-level vaccination and intent, Bivalent Booster Uptake and Intention]
+    indicator: [received a vaccination, Received updated bivalent booster dose (among adults who completed primary series)]
```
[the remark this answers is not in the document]
eschrom: That's not a terrible problem, since we already know the relevant config entries to extract what we want.

iup/__init__.py
eschrom: Fixed (using the config parameter)

iup/models.py
eschrom: Fixed.

scripts/eval.py
eschrom: I had resisted this at first, thinking that it's easy to forget "season" since the reason we group by "season" is a little different than the reason we would consider grouping by "demographic" or "geography" in the future. But perhaps the solution to that concern is, "well then just don't forget." This is in the config now.

iup/models.py
eschrom: `pl.is_between()`  -  Fixed.

eschrom:
All of @swo's comments are addressed. Although the autoregressive model does not perform great, I have checked that it now runs without errors (and, importantly, without logical flaws, I believe!) for both covid and flu. This is ready for you when you have the time, @fuhan!

By the way, merging this will close #88.

Fuhan-Yang (commented):
Thanks Ed for debugging the pipeline! My comments below are mostly minor. While the current version is forecasting between 2024-02-03 and 2024-04-30, I run the pipeline given monthly forecast starts as [2023-09-01, 2023-10-01, 2023-11-01, 2023-12-01, 2024-01-01]. There are errors when :
- Forecast start is 2023-09-01: Large negative values (~-1e9) showed up in the forecast between 2023-09-01 and 2024-04-30. My institution is that the negativity accumulates as the prediction is used as the predictor in next time step. Maybe variable transformation is needed (like the truncated normal we talked before)? l opened this as an issue
- Forecast start is 2023-10-01: Seems like a predicted cumulative projection is empty, maybe this is related to rollout dates?

Other forecast starts run smoothly.

iup/__init__.py
```diff
@@ -136,7 +136,7 @@ def to_cumulative(
         out = self.with_columns(estimate=pl.col("estimate").cum_sum().over(group_cols))
 
         if last_cumulative is not None:
```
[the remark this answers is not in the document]
eschrom: I'm not sure I understand - `last_cumulative` is either a `pl.DataFrame` with 1+ rows, or else it is not provided and defaults to `None`. I don't see a case where an empty data frame would be passed to `to_cumulative`

scripts/preprocess.py
```diff
@@ -16,14 +16,26 @@ def preprocess(
...
+        .collect()
+    )
 
     # Ensure that the desired grouping factors are found in all data sets
     assert set(data.columns).issuperset(groups)
 
     # Insert rollout dates into the data
-    return iup.CumulativeUptakeData(data).insert_rollouts(rollouts, groups)
+    return iup.CumulativeUptakeData(
+        data.insert_rollouts(rollouts, groups).with_columns(
```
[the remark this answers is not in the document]
eschrom: I have no idea! It is convenient, though. I am finding that there are several aspects of .yaml files that seem to work almost like magic.

scripts/preprocess.py
```diff
@@ -16,14 +16,26 @@ def preprocess(
     groups: List[str],
```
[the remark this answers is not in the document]
eschrom: Rollout is defined as the day the vaccine was first available, when uptake must be 0, so I don't think that the `rollouts` config entry or the `insert_rollouts` function would be particularly confusing to users. 

The date range of data used for model training is always "everything before the forecast date." I don't think users will struggle with this, because it's the most natural scenario. To the contrary, _excluding_ some data that would have been available for training as of the forecast date is what would demand an explanation.

iup/__init__.py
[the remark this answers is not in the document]
Fuhan-Yang: I see, as long as there is no such case that `last_cumulative` is an empty data frame, this should be fine

eschrom:
I will re-requesting your review, @Fuhan, because the two biggest comments are addressed in the threads for #88 and #124. Some of the more minor points I tried to address, but I'm not sure I understood what you were saying correctly. If my explanations don't help, let me know!

scripts/preprocess.py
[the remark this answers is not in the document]
Fuhan-Yang: In this way, the users cannot specify the amount of training data from config, they will only know from reading the parquet. Maybe this is ok? One issue I can think about is that we cannot compare the models across regions or demographics in the future, due to discrepancies of training data. But as long as we ensure multiple models are built under the same filter, model comparison is still fair.

Fuhan-Yang:
I'm thinking about what we assume the users should know when using the package. Say if an user want to forecast the flu vaccine in season 2024/2025, they need to know the certain rollout dates to make sure the training data is correctly selected. Then, as you mentioned, they need to know that the forecast start and end dates exactly exist in the test data. Maybe we can add the option that if the rollout/forecast start/forecast end does not match the data, we can use the nearest data point for whichever purpose? But this seems to be the next PR.

Fuhan-Yang:
I can edit the postprocess file to only show the data that has been forecasted. I can put it in another PR.

scripts/preprocess.py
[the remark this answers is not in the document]
eschrom: My point is that the user doesn't need to know the amount of training data in absolute terms - they just need to know that all data (that matches their filters) before the forecast date was used for training.

Once multiple models are specified, they will indeed all be built with identically filtered data, because the filtering happens just once before any/all models are fit. 

Is your concern that there may be different amounts of data per region or demographic? E.g. what if there are more data points for Iowa than for Connecticut? I think the number of data points is always or almost always the same across regions or demographics. And even if it wasn't, that's okay - we are interested in comparing Model 1 vs. Model 2 on Iowa and Model 1 vs. Model 2 on Connecticut, not Model 1 on Iowa vs. Connecticut and Model 2 on Iowa vs. Connecticut.

Fuhan-Yang (approved):
We still need to resolve the issues of rollout dates and forecast start and end don't match the user's input in config, and change the plot in postprocess. But I think this PR already done its task!

eschrom:
It is a lot to ask the user to know the rollout date for every past season of flu. But we will soon refactor our models to infer rollout date as a parameter rather than accept it as data. So this inconvenience should go away soon, at least for some types of models.

Also, I've already done a few of the things you mentioned:
- The user does not need to know the exact dates in the test data. If evaluation is being done, `build_scaffold` in `models.py` (around line 347) now chooses the dates in the test data nearest to those specified in the config, just as you suggested!
- The postprocess file only plots within the season that forecasts were performed, discarding prior and future seasons. This is around line 38 in `postprocess.py`

PR #131, opened 2025-03-07
Standard deviation of estimate in preprocessed data

eschrom:
In anticipation of refactored models that require the estimated error in observed uptake, this PR adds the standard deviation of observed uptake as a column in the preprocessed data. 

I don't know how the upper and lower bounds of the "95% confidence intervals" in the NIS data are calculated. But with a truncated normal distribution in mind to describe the observation error, I have taken the width of the reported 95% confidence interval, divided by 2, and then divided by ~1.96 to estimate a standard deviation of sorts for the observed uptake.


eschrom:
Oops - botched my branching structure. Gonna start over.

Issue #28, opened 2024-10-25
Data objects need (optional) confidence intervals


eschrom:
Cumulative and incident uptake data objects (whether they represent input data or model projections) should have not only point estimates, but also columns for upper and lower bounds of credible intervals.

swo:
- This will become relevant when we do Bayesian models that need input data error distributions
- We probably want some validation but not super-hard because the data API should take care of that mostly

eschrom:
Completed with #132

PR #132, opened 2025-03-07
Std dev of estimate as a data column

eschrom:
In anticipation of refactored models that require the estimated error in observed uptake, this PR adds the standard deviation of observed uptake as a column in the preprocessed data.

I don't know how the upper and lower bounds of the "95% confidence intervals" in the NIS data are calculated. But with a truncated normal distribution in mind to describe the observation error, I have taken the width of the reported 95% confidence interval, divided by 2, and then divided by ~1.96 to estimate a standard deviation of sorts for the observed uptake.

### DIFF
--- a/scripts/config_template.yaml
+++ b/scripts/config_template.yaml
@@ -13,7 +13,7 @@ data:
     indicator: [received a vaccination, Received updated bivalent booster dose (among adults who completed primary series)]
     time_type: [week]
   # keep only these data columns
-  keep: [estimate, time_end]
+  keep: [estimate, time_end, lci, uci]
   # use these columns as grouping factors. Almost always use at least "season", but use None if there are truly none.
   groups:
 
--- a/scripts/preprocess.py
+++ b/scripts/preprocess.py
@@ -29,8 +29,10 @@ def preprocess(
                 season_start_month=season_start_month,
                 season_start_day=season_start_day,
             ),
-            sdev=(pl.col("uci") - pl.col("lci")) / (2 * st.norm.ppf(0.975, 0, 1)),
+            uci=pl.col("uci") - pl.col("estimate"),
+            lci=pl.col("estimate") - pl.col("lci"),
         )
+        .with_columns(sdev=pl.max_horizontal("uci", "lci") / st.norm.ppf(0.975, 0, 1))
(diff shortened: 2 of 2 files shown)


swo:
It might be good to check that the 95 intervals are symmetrical. I'm fairly certain that they are, but this would double-check that it makes sense to assume that they are a normal distribution (that comes from some approximation of the error distribution from the complex survey design)

swo:
Does this resolve #28 ?

eschrom:
Why yes it does resolve #28. I hadn't even realized.

Spot checking reveals that the 95% confidence intervals reported in the NIS data are usually but not always symmetric. When they are asymmetric, it's sometimes because the estimate is too close to 0, so it seems the lower CI bound has just been reset to 0. Other times, the estimate isn't that close to 0, but the CI isn't quite symmetric anyway. 

Do we have any information on how these CIs were originally calculated?

eschrom:
I have changed my method of estimating the standard deviation of observed uptake: I now take the maximum of (upper bound - estimate) and (estimate - lower bound) and divide by ~1.96. This is the same as before when the reported 95% confidence interval is symmetric, but it also covers the case when the reported 95% confidence interval is asymmetric due to truncation at 0 or 1.

afmagee42:
> I have changed my method of estimating the standard deviation of observed uptake: I now take the maximum of (upper bound - estimate) and (estimate - lower bound) and divide by ~1.96. This is the same as before when the reported 95% confidence interval is symmetric, but it also covers the case when the reported 95% confidence interval is asymmetric due to truncation at 0 or 1.

Seems reasonable, but I also agree it would be nice to know what went into their creation in the first place.

Fuhan-Yang (approved):
[There ](https://www.cdc.gov/fluvaxview/dashboard/children-vaccination-coverage.html)is a description about 95% CI: "Note: Confidence Intervals (CI) describe the level of uncertainty around an estimate because a sample was taken. 95% CIs represent the range of values that would result if the data collection had been repeated many times. For a 95% confidence interval, if the sampling method is repeated many times, the true value would fall within this interval at least 95% of the time. Wider CIs reflect larger random error in estimates resulting from survey sampling. " Based on that, I think the method here is reasonable!

PR #133, opened 2025-03-07
Refactor forecast script

swo:
- Change the variable names to `config_model` (dictionary of config stuff), `model_name` (a string), `model_class` (the subclass of `UptakeModel`), and `fit_model`.
  - Confusingly, the code currently refers to the model class as `model _name` and the string-name of the model with `model["name"]`.
  - Note we don't actually ever have just `model` here.
- Explicitly pass config stuff to `run_forecast()`
  - Currently, the code refers to `config` in `run_forecast()`, even though `config` isn't passed as a parameter, which means it's counting on it being in the outer namespace as a closure, which could be very confusing
  - Now, `run_forecast()` doesn't know anything about `config` or its structure; it just gets passed the values it wants

### DIFF
--- a/scripts/forecast.py
+++ b/scripts/forecast.py
@@ -1,4 +1,5 @@
 import argparse
+from typing import Any, Type
 
 import polars as pl
 import yaml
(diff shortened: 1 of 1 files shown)


eschrom (approved):
All makes sense to me.

PR #135, opened 2025-03-07
Refactor forecast script (alternative)

swo:
This is another approach to #133 

- Do the same renaming as in #133 
- Rather than looping over models, setting up augmented data, then looping over dates, and calling the one-forecast function, instead loop over models & forecast dates and call the one-forecast function.
  - Con: you augment the data multiple times
  - Pro: there are fewer interfaces, and only one layer of calls.
  - Compared to #133 , you don't end up with as desirably-dumb one-forecast function, but it means that you don't need to define another function interface

I encourage #133; this one is more of a suggestion for consideration; feel free to close if it's not interesting

### DIFF
--- a/scripts/forecast.py
+++ b/scripts/forecast.py
@@ -1,4 +1,5 @@
 import argparse
+from typing import Any, Type
 
 import polars as pl
 import yaml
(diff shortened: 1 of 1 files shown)


eschrom:
#133 is more intuitive to me, so that's the one I would like to use, as well.

PR #141, opened 2025-03-14
Refactor Hill Model

eschrom:
The Hill model is refactored to handle arbitrary grouping factors (including but not limited to season), separate observation from latent true uptake, etc. See docs/model_details.md for the full description of this model.

This is currently a work in progress. In particular, I am running into `Segmentation fault (core dumped)` when I run the model on flu data across 14 seasons and 54 states/jurisdictions. I did not expect this to be a large-enough data set or a complex-enough model to crash my CPU, so more investigation must follow.

### DIFF
--- a/docs/model_details.md
+++ b/docs/model_details.md
@@ -89,7 +89,7 @@ At a high level, the Hill model is structured as follows:
 \end{align*}
 ```
 
-More details on the Hill model are given below.
+For the Hill model, $t$ is rescaled by dividing by 365, so that $t$ represents the proportion of a season elapsed. This is not a problem because $t$ need not be an integer here, and it helps keep all parameters on a similar scale. More details on the Hill model are given below.
 
 ## Observation Layer
 
@@ -122,10 +122,10 @@ and similarly for $H$.
 
 ```math
 \begin{align*}
-&A \sim \text{Beta}(\text{shape1 = }2.4, \text{ shape2 = }3.6) \\
-&H \sim \text{Gamma}(\text{shape = }100.0, \text{ rate = }1.0) \\
-&\sigma_{A_{G_i}} \sim \text{Exponential}(\text{mean = }0.1) ~\forall~i~\text{ in } 1, ..., I \\
-&\sigma_{H_{G_i}} \sim \text{Exponential}(\text{mean = }5.0) ~\forall~i~\text{ in } 1, ..., I \\
-&n \sim \text{Gamma}(\text{shape = }2.0, \text{ rate = }1.0) \\
+&A \sim \text{Beta}(\text{shape1 = }15.0, \text{ shape2 = }20.0) \\
+&H \sim \text{Beta}(\text{shape1 = }25.0, \text{ shape2 = }50.0) \\
+&\sigma_{A_{G_i}} \sim \text{Exponential}(\text{mean = }40.0) ~\forall~i~\text{ in } 1, ..., I \\
+&\sigma_{H_{G_i}} \sim \text{Exponential}(\text{mean = }40.0) ~\forall~i~\text{ in } 1, ..., I \\
+&n \sim \text{Gamma}(\text{shape = }20.0, \text{ rate = }5.0) \\
 \end{align*}
 ```
--- a/docs/model_details.md
+++ b/docs/model_details.md
@@ -122,9 +122,10 @@ and similarly for $H$.
 
 ```math
 \begin{align*}
-&A \sim \text{Beta}(\text{mean = }0.4, \text{ shape = }0.2) \\
-&H \sim \text{Gamma}(\text{shape = }100, \text{ scale = }1) \\
+&A \sim \text{Beta}(\text{shape1 = }2.4, \text{ shape2 = }3.6) \\
+&H \sim \text{Gamma}(\text{shape = }100.0, \text{ rate = }1.0) \\
 &\sigma_{A_{G_i}} \sim \text{Exponential}(\text{mean = }0.1) ~\forall~i~\text{ in } 1, ..., I \\
-&\sigma_{H_{G_i}} \sim \text{Exponential}(\text{mean = }0.1) ~\forall~i~\text{ in } 1, ..., I \\
+&\sigma_{H_{G_i}} \sim \text{Exponential}(\text{mean = }5.0) ~\forall~i~\text{ in } 1, ..., I \\
+&n \sim \text{Gamma}(\text{shape = }2.0, \text{ rate = }1.0) \\
 \end{align*}
 ```
--- a/docs/model_details.md
+++ b/docs/model_details.md
@@ -124,8 +124,8 @@ and similarly for $H$.
 \begin{align*}
 &A \sim \text{Beta}(\text{shape1 = }15.0, \text{ shape2 = }20.0) \\
 &H \sim \text{Beta}(\text{shape1 = }25.0, \text{ shape2 = }50.0) \\
-&\sigma_{A_{G_i}} \sim \text{Exponential}(\text{mean = }40.0) ~\forall~i~\text{ in } 1, ..., I \\
-&\sigma_{H_{G_i}} \sim \text{Exponential}(\text{mean = }40.0) ~\forall~i~\text{ in } 1, ..., I \\
+&\sigma_{A_{G_i}} \sim \text{Exponential}(\text{rate = }40.0) ~\forall~i~\text{ in } 1, ..., I \\
+&\sigma_{H_{G_i}} \sim \text{Exponential}(\text{rate = }40.0) ~\forall~i~\text{ in } 1, ..., I \\
 &n \sim \text{Gamma}(\text{shape = }20.0, \text{ rate = }5.0) \\
 \end{align*}
 ```
--- a/iup/models.py
+++ b/iup/models.py
@@ -8,7 +8,7 @@ import numpyro
 import numpyro.distributions as dist
 import polars as pl
 from jax import random
-from numpyro.infer import MCMC, NUTS, Predictive
+from numpyro.infer import MCMC, NUTS, Predictive, init_to_sample
 from typing_extensions import Self
 
 import iup.utils
@@ -657,12 +657,12 @@ class HillModel(UptakeModel):
         group_names=None,
         A_shape1=15.0,
         A_shape2=20.0,
-        A_sig=5.0,
-        H_shape=100.0,
-        H_rate=1.0,
-        H_sig=0.5,
-        n_shape=2.0,
-        n_rate=1.0,
+        A_sig=20.0,
+        H_shape1=25.0,
+        H_shape2=50.0,
+        H_sig=20.0,
+        n_shape=20.0,
+        n_rate=0.2,
     ):
         """
         Fit a Hill model on training data.
@@ -689,7 +689,7 @@ class HillModel(UptakeModel):
         """
         # Sample the overall average value for each Hill function parameter
         A = numpyro.sample("A", dist.Beta(A_shape1, A_shape2))
-        H = numpyro.sample("H", dist.Gamma(H_shape, H_rate))
+        H = numpyro.sample("H", dist.Beta(H_shape1, H_shape2))
         n = numpyro.sample("n", dist.Gamma(n_shape, n_rate))
         if groups is not None and group_names is not None:
             num_data_points = groups.shape[0]
@@ -744,15 +744,6 @@ class HillModel(UptakeModel):
             # Calculate the postulated latent true uptake given the time elapsed each data
             # point, accounting for the ultimate A and H values dictated by grouping factors
             mu = A_tot * (elapsed**n) / (H_tot**n + elapsed**n)
-            print(A_sig)
-            print(A_sigs)
-            print(H_sig)
-            print(H_sigs)
-            print("BREAK")
-            # print(jnp.sum(jnp.isnan(mu))) # Both A and H can be less than 0, but only H < 0 causes NaN
-            # print(jnp.sum(A_tot < 0))     # Four replicates are printed because there are 4 chains!
-            # print(jnp.sum(H_tot < 0))
-            # print("BREAK")
         else:
             # Without grouping factors, use the same A and H for every data point
             mu = A * (elapsed**n) / (H**n + elapsed**n)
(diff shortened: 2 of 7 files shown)


eschrom:
The refactored Hill model is ready for review! See the [model details doc](https://github.com/CDCgov/cfa-immunization-uptake-projection/blob/main/docs/model_details.md) for a description of the intended model structure (and note that there are updates to that doc in this PR).

I have tested the model using season and state as grouping factors: it runs successfully and produces group-specific forecasts in the right ballpark!

That said, I would still like to direct your attention to several caveats and questions:
- A few changes in this PR (e.g. adjustments to scaffold building) might cause the linear incident uptake model to break. Let's not worry about that, since the linear incident uptake model will be replaced by a refined autoregressive model soon anyway.
- The refactored Hill model accommodates arbitrary grouping factors (each with arbitrary levels) without using plates! Is my approach sensible and general? If so, should we document this elsewhere for use in other projects?
- My approach requires a little bookkeeping of numeric codes for group levels, to insure that group levels are interpretted identically between the tasks of fitting and prediction. This bookkeeping currently appears in `HillModel.fit()` and `HillModel.predict()`. Should this bookkeeping be outsourced to separate, modular functions, perhaps in `utils`?
- The accommodation of arbitrary grouping factors in this PR will cause downstream evaluation and plotting to break. @Fuhan-Yang, would you like to address that in a separate future PR, or would you like me to try to address that here?
- Despite rescaling some parameters to keep them all within the same order of magnitude and choosing fairly targeted priors, MCMC is still _very_ slow (~1-2 it/s, for both warmup and sampling). Have I missed an obvious inefficiency or misspecification? Or do I just have unrealistic expectations of speed when 2 different parameters take on different values for each of 14 seasons and 54 states?

eschrom:
Here is a diagram to explain the bookkeeping on the levels of each grouping factor:
image

afmagee42:
> * Despite rescaling some parameters to keep them all within the same order of magnitude and choosing fairly targeted priors, MCMC is still _very_ slow (~1-2 it/s, for both warmup and sampling). Have I missed an obvious inefficiency or misspecification? Or do I just have unrealistic expectations of speed when 2 different parameters take on different values for each of 14 seasons and 54 states?

There are many potential causes of slow MCMC. I'm thinking this might be a misspecification problem, because running this model, I'm seeing this:

warmup:   5%|▍        | 95/2000 [00:15<05:09,  1.16it/s, 1023 steps of size 4.36e-04. acc. prob=0.74]

This tells me that the MCMC is very unhappy: its acceptance probability is nowhere near where it wants to be and it's taking a ton of steps (presumably to try to get there).

(Gelman's) Folk Theorem interpretation would be that the MCMC is trying to tell us the model (encompassing everything from the choice of fixed hyperparameters through prior distribution families to the likelihood) is not a good fit.

afmagee42:
> * The standard deviation for the truncated normal likelihood cannot be 0. This is a problem in two contexts: 1) a handful of the data points report an observed standard deviation of zero, and 2) there is no observed standard deviation when performing prediction. I didn't know what to do in either case ... so I just shamefully declared a standard deviation of 0.0001 in both cases. What are the right ways to handle these cases?

I make no claim on "right," but "more defensible" maybe.
- If the SDs don't show any obvious trends throughout a season, you could:
  - Replacing small values with
    - Some small quantile of the non-zero SDs
    - Assume that 0 is a rounding error and that the variance is basically the largest possible value that would get rounded to 0. So, if all SDs are reported to the 4th decimal place, take 0 to mean 0.00005.
  - Predict using the (square root  of the) mean (variance (variances are additive, not SDs)) would seem better. Or the median (which you can take on the variance scale, quantiles are preserved by monotonic transforms).
- Trend or not, you can also make the model do the work: model the SDs (well, model the _variances_ and then transform them) somehow. That will give us something for predictions too.

iup/models.py
```diff
@@ -650,58 +651,96 @@ def __init__(self, seed: int):
...
-                "H_s",
-                dist.TruncatedNormal(H, sig_H, low=H_low, high=H_high),
-                sample_shape=(len(np.unique(season)),),
+            H_devs = numpyro.sample(
+                "H_devs", dist.Normal(0, 1), sample_shape=(sum(num_group_levels),)
             )
-            mu = A_s[season] * (elapsed**n) / (H_s[season] ** n + elapsed**n)
+            # Scale the centered deviation for each level of each grouping factor
+            # by the characteristic spread for each grouping factor
+            A_devs = A_devs * jnp.repeat(
```
afmagee42, line 716: I'm not sure how JAX handles reassignment of variable names, but I generally try to avoid it. Here, we could just slot in the `numpyro.sample()` statement and avoid the first declaration entirely (the non-expanded `_devs` will still be tracked in the output by dint of being declared inside a sample `.sample()` statement.

iup/models.py
```diff
@@ -650,58 +651,96 @@ def __init__(self, seed: int):
...
+            H_devs = H_devs * jnp.repeat(
+                H_sigs,
+                jnp.array(num_group_levels),
+                total_repeat_length=sum(num_group_levels),
+            )
+            # Across all data points, look up the A and H deviations due to the
+            # grouping factors. Sum across grouping factors and include the overall
+            # average A and H to get the final total A and H for each datum
+            A_tot = jnp.sum(A_devs[groups], axis=1) + A
+            H_tot = jnp.sum(H_devs[groups], axis=1) + H
```
afmagee42, line 730: My numpy indexing is weak, but are we sure this is doing what we want it to do?

iup/models.py
```diff
@@ -766,19 +830,54 @@ def fit(
...
+        # and the number of levels for each grouping factor. These will be
+        # useful to have saved as model attributes during prediction.
+        if groups is not None:
+            group_codes = data.select(groups)
+            self.num_group_factors = group_codes.shape[1]
+            self.value_to_index = {}
+            for i in range(self.num_group_factors):
+                col_name = group_codes.columns[i]
+                unique_values = (
+                    group_codes.select(col_name).unique().to_series().to_list()
```
afmagee42, line 845: Is the `.to_series()` necessary when it's being `.to_list()`ed anyways?

iup/models.py
```diff
@@ -766,19 +830,54 @@ def fit(
...
-            season = data["season"].to_numpy()
-            unique_seasons = np.unique(season)
-            season_to_index = {s: i for i, s in enumerate(unique_seasons)}
-            season = np.array([season_to_index[s] for s in season])
-            self.season = season
+        # Tranform the levels of the grouping factors into numeric codes
+        # Make a dictionary of dictionaries that map levels to numeric codes
+        # Save this as a model attribute, along with the number of grouping factors
+        # and the number of levels for each grouping factor. These will be
+        # useful to have saved as model attributes during prediction.
```
afmagee42, line 837: We definitely want to write "give me the groups as integers" just once, and then call it elsewhere.

It's a general notion, and you'll want this in the refactored linear incident uptake model as well as these two places, which crosses the magical number of three at which point refactoring to pull it out is advised.

iup/models.py
```diff
@@ -766,19 +830,54 @@ def fit(
...
+                unique_values = (
+                    group_codes.select(col_name).unique().to_series().to_list()
+                )
+                self.value_to_index[col_name] = {
+                    v: j for j, v in enumerate(unique_values)
+                }
+                group_codes = group_codes.with_columns(
+                    pl.col(col_name)
+                    .replace(self.value_to_index[col_name])
+                    .cast(pl.Int8)
```
afmagee42, line 853: Might as well cast these to unsigned ints, since they're going to have to be positive to be used as indexes?

afmagee42 (approved):
I've tried to point out a few places that could be responsible for either slow gradient computations or model misspecifcation. But I can't be sure how of the probability of success.

I'm also starting to wonder if a matrix-based one-hot-encoding of the group factors is worth pursuing. It would require a decently-sized matrix, but it would potentially let us get both A and H done via the same matrix multiplications rather than separately `jnp.repeat`ing them.

eschrom:
> (Gelman's) Folk Theorem interpretation would be that the MCMC is trying to tell us the model (encompassing everything from the choice of fixed hyperparameters through prior distribution families to the likelihood) is not a good fit.

I ran the model without any grouping factors, to fit the three Hill function parameters sans hierarchy. Here are the 2022/2023 data vs. the model trained on all seasons from 2009/2010 through 2022/2023:

image

One reason for the dodgy fit is because the curve represents the average of all seasons, whereas the points represent just one season. But another reason is that after ~February, the Hill function asymptotes whereas real uptake continues to creep up. 

This could be a clue that the Hill function is just not a good fit. I doubt that's the whole story behind the super-slow MCMC with grouping factors included (for the overall model, MCMC hit a crisp 653it/s), but I thought it worth documenting here.

docs/model_details.md
```diff
@@ -122,9 +122,10 @@ and similarly for $H$.
...
 ```math
 \begin{align*}
-&A \sim \text{Beta}(\text{mean = }0.4, \text{ shape = }0.2) \\
-&H \sim \text{Gamma}(\text{shape = }100, \text{ scale = }1) \\
-&\sigma_{A_{G_i}} \sim \text{Exponential}(\text{mean = }0.1) ~\forall~i~\text{ in } 1, ..., I \\
-&\sigma_{H_{G_i}} \sim \text{Exponential}(\text{mean = }0.1) ~\forall~i~\text{ in } 1, ..., I \\
+&A \sim \text{Beta}(\text{shape1 = }15.0, \text{ shape2 = }20.0) \\
+&H \sim \text{Beta}(\text{shape1 = }25.0, \text{ shape2 = }50.0) \\
+&\sigma_{A_{G_i}} \sim \text{Exponential}(\text{mean = }40.0) ~\forall~i~\text{ in } 1, ..., I \\
+&\sigma_{H_{G_i}} \sim \text{Exponential}(\text{mean = }40.0) ~\forall~i~\text{ in } 1, ..., I \\
```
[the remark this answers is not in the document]
eschrom: Just a typo in doc, now fixed. The rate should be 40 and the mean 0.025.

iup/models.py
```diff
@@ -650,58 +651,96 @@ def __init__(self, seed: int):
...
+                "A_sigs", dist.Exponential(A_sig), sample_shape=(num_group_factors,)
+            )
+            H_sigs = numpyro.sample(
+                "H_sigs", dist.Exponential(H_sig), sample_shape=(num_group_factors,)
+            )
+            # Draw a sample of the centered deviation from the overall average
+            # A and H for each level of each grouping factor
+            A_devs = numpyro.sample(
+                "A_devs", dist.Normal(0, 1), sample_shape=(sum(num_group_levels),)
             )
```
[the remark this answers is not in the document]
eschrom: At first I assumed JAX did support ragged arrays and got burned!

iup/models.py
eschrom: I didn't realize the non-expanded `_devs` would still be tracked, but seeing as they are named by `.sample()`, this makes total sense. Good suggestion!

afmagee42:
> This could be a clue that the Hill function is just not a good fit. I doubt that's the whole story behind the super-slow MCMC with grouping factors included (for the overall model, MCMC hit a crisp 653it/s), but I thought it worth documenting here.

Agreed, there are degrees of badness and plenty bad models fit easily enough. This does suggest some experiments, though:
- Season-only grouping (pooled data across all states for all seasons or only one state)
- State-only grouping (for one season or all seasons pooled)
- Half the states, half the seasons (or some other intermediate form)

iup/models.py
```diff
@@ -650,58 +651,96 @@ def __init__(self, seed: int):
...
-                sample_shape=(len(np.unique(season)),),
+            H_devs = numpyro.sample(
+                "H_devs", dist.Normal(0, 1), sample_shape=(sum(num_group_levels),)
             )
-            mu = A_s[season] * (elapsed**n) / (H_s[season] ** n + elapsed**n)
+            # Scale the centered deviation for each level of each grouping factor
+            # by the characteristic spread for each grouping factor
+            A_devs = A_devs * jnp.repeat(
+                A_sigs,
+                jnp.array(num_group_levels),
```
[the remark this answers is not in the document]
eschrom: I tested this out and numpy does work in place of jax.numpy - happy to make the switch.

iup/models.py
eschrom: Yes sir. It took a little experimenting to convince myself, but in the end I was impressed by how convenient this turns out to be. Here is a little demo:
```
import numpy as np
A_devs = np.array([-0.2, -0.1, 0.1, 0.2])
A_devs
#     array([-0.2, -0.1,  0.1,  0.2])
groups = np.array([[0,2],[0,3],[1,2],[1,3]])
groups
#     array([[0, 2],
#            [0, 3],
#            [1, 2],
#            [1, 3]])
A_devs[groups]
#     array([[-0.2,  0.1],
#            [-0.2,  0.2],
#            [-0.1,  0.1],
#            [-0.1,  0.2]])
np.sum(A_devs[groups], axis=1) + 4
#     array([3.9, 4. , 4. , 4.1])
```

iup/models.py
eschrom: Weirdly yes - you cannot directly `.to_list()` a polars data frame column.

iup/models.py
eschrom: Fair enough - done.

eschrom:
> * If the SDs don't show any obvious trends throughout a season, you could:
>   
>   * Replacing small values with
>     
>     * Some small quantile of the non-zero SDs
>     * Assume that 0 is a rounding error and that the variance is basically the largest possible value that would get rounded to 0. So, if all SDs are reported to the 4th decimal place, take 0 to mean 0.00005.
>   * Predict using the (square root  of the) mean (variance (variances are additive, not SDs)) would seem better. Or the median (which you can take on the variance scale, quantiles are preserved by monotonic transforms).
> * Trend or not, you can also make the model do the work: model the SDs (well, model the _variances_ and then transform them) somehow. That will give us something for predictions too.

SDs are rounded to the nearest 0.000001 (millionth), so I've replaced those with 0.0000005, preferring not to add complication to the model until the current slowness is solved.

eschrom:
Good news! And an embarrassing admission...

My replacement of the SDs with a small number was only working if no observed SDs were given at all. For the >80 instances of SD = 0 in the observed data, SD was not being replaced by a small number. Rather than complaining, numpyro was just grinding through an impossible search for parameter values that could hit all these >80 "error-less" observations perfectly.

This was the cause of the slow speed. Now that I have corrected my approach to replace these >80 SDs of 0 with a small number, numpyro rips through 1000 warmups + 1000 samples in ~5s, even with 14 seasons and 54 states.

PR #142, opened 2025-03-18
Validate incident uptake

### LINKED ISSUE #68: Validate incident uptake
Incident uptake is also a proportion, although we accept the fact that it can sometimes be reported as a negative. Should we implement a similar assertion that each "estimate" in an IncidentUptakeData is between -1.0 and 1.0?

_Originally posted by @eschrom in https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/65#discussion_r1868185533_

swo:
- Assert that incident uptakes are also proportions
- Rework tests to match, by dividing all the uptakes by 10

Resolves #68

### DIFF
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -82,6 +82,24 @@ class UptakeData(Data):
 
 
 class IncidentUptakeData(UptakeData):
+    def validate(self):
+        # same validations as UptakeData
+        super().validate()
+        # and also require that uptake be a proportion. Incident uptakes can be negative because
+        # of corrections or data errors, but the biggest jumps possible are from 0% up to 100%
+        # (i.e., +1.0) or from 100% down to 0% (i.e., -1.0). We do not do further validation, e.g.
+        # to check that cumulative uptake is always within 0% to 100%
+        if not self["estimate"].is_between(-1.0, 1.0).all():
+            bad_values = (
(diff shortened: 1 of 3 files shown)


iup/__init__.py
```diff
@@ -82,6 +82,24 @@ def split_train_test(
...
 class IncidentUptakeData(UptakeData):
+    def validate(self):
+        # same validations as UptakeData
+        super().validate()
+        # and also require that uptake be a proportion. Incident uptakes can be negative because
+        # of corrections or data errors, but the biggest jumps possible are from 0% up to 100%
+        # (i.e., +1.0) or from 100% down to 0% (i.e., -1.0). We do not do further validation, e.g.
+        # to check that cumulative uptake is always within 0% to 100%
+        if not self["estimate"].is_between(-1.0, 1.0).all():
+            bad_values = (
```
eschrom, line 93: I'm not sure whether it's more useful to report (as here) what the illegal values are, or (alternatively) where they occur in the data. Either way, this is probably more sophisticated than we need - the validation alone probably suffices. Looks good to me!

eschrom (approved):
Looks great!

iup/__init__.py
swo: I wrote that in when the tests failed and I couldn't quite figure out why :)

PR #143, opened 2025-03-20
Print posterior samples to a file

eschrom:
This PR modifies the forecast.py script to write a parquet file of the posterior samples for each model x forecast date, in addition to the parquet file of the forecast itself for each model x forecast date.


eschrom:
To write parquet files of the posterior samples, I started by `poetry add`ing arviz and pandas - arviz for handling posterior samples and pandas because arviz only exports directly to pandas, not polars. Most of the work is in scripts/forecast.py, where we must now return both forecasts as well as posterior samples.

Note that forecasts from different models and forecast dates are concatenated into a single data frame, written to a single parquet. But this doesn't work for posterior samples, because posterior samples from different models contain different parameters, and thus have different columns. So posterior samples from all forecast dates within a model are concatenated into a data frame, but these data frames are listed across models and exported to separate .parquet files.

Finally, notice that the strategy to handle arbitrary grouping factors and their levels from #141 means that group-specific parameters are auto-named using numeric indices (e.g. "A_Devs_17" rather than "A_Devs_Alaska"). It takes a little work to swap the correct strings in for these numeric codes.

Fuhan-Yang (approved):
This looks good to me! This is one of tasks I plan to do, which are generating posterior and diagnostic plots. Since there will be multiple outputs from `fit_model`: posterior, model summary, diagnostic plots, I plan to decompose `run_forecast` into `fit_model` and `run_forecast` and include the locations of these outputs in Makefile. This is a good start!

Issue #128, opened 2025-03-05
Refactor the standardize utility


eschrom:
Make a cleaner version that works for numpy or polars inputs, handles 0 sd correctly, and isn't so confusing. See suggestions: 

    loc = mn if mn is not None else x.mean()
    scale = sd if sd is not None else x.std()
    if scale == 0.0:
        return x * 0.0
    return (x - loc) / scale

But,

> If you want something that behaves the same way for `np.ndarray` as for polars things, you'll need `np.nanmean` and `np.nanstd` with a switch to catch numpy arrays, rather than polars things (unless I'm spectacularly wrong and there's a big Expressions/Series gap here of which I'm not currently aware).
> 
> But then you'll also get the fun that `np.nanstd` doesn't agree with `pl.col.std`. I _think_ polars is using the Bessel correction by default and numpy isn't.

eschrom:
Closed with #148

PR #149, opened 2025-03-31
More careful standardization

eschrom:
The standardization utility now handles NaNs consistently and does not apply the Bessel correction, regardless of whether the input is a numpy ndarray or a polars expression. I tried to streamline the function to look cleaner, but lurking around every corner was some operation or conditional that required separating the ndarray vs. expression cases anyway. I don't think this is worth any more effort.

This will address #128


Fuhan-Yang (approved):
Looks good to me!

PR #153, opened 2025-04-03
Posterior predictive check plots

eschrom:
This requires no code review - I've merely replaced the outdated stuff in my scratchwork script ("model_demo.r") with some rudimentary posterior predictive plots. There is plenty to improve, but I think this shows we are on the right track.

I load the flu data and the logistic+linear model posterior that this repo produces as a .parquet (see #143). For each season from 2009/2010 through 2021/2022, I plot the mean predicted latent true uptake vs. the data across states. (The 95% credible interval on the latent true uptake is also included, but it is usually quite thin.)

Note that the model was fit with 1 chain (not enough) using 500 warmups (not enough) and 500 samples (not enough), so there is room for immediate improvement.

Here's an example of the 2016-2017 season.

image

Since this requires no code review, we can just merge it, but I wanted everyone to see where we are currently.

cc @Fuhan-Yang @afmagee42 @swo

### DIFF
--- a/scripts/model_demo.r
+++ b/scripts/model_demo.r
@@ -1,4 +1,4 @@
-# This is a demo of our candidate model structures
+# Scratchwork to test and evaluate models
 
 # Relevant data sets from the NIS API are:
 #    ksfb-ug5d: recent covid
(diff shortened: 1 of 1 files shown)


Fuhan-Yang (approved):
The plot looks great!

Issue #161, opened 2025-04-09
Looping single dictionary with multiple entries versus a list with multiple one-entry dictionary


Fuhan-Yang:
As I've already mentioned, this type of loop would be cleaner if `sel_model_dicts` was just a single dictionary that's a subset of the `models` input dictionary. The same applies to `diagnostic_table` below.

_Originally posted by @eschrom in https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/150#discussion_r2035657787_

eschrom:
Addressed as of #162

PR #162, opened 2025-04-10
Dictionary comprehension for a slightly cleaner diagnostics script

eschrom:
Dictionary comprehension makes the `diagnostics.py` script slightly cleaner. This addresses #161


Fuhan-Yang (approved):
This is much clearer! Thanks Ed!

PR #164, opened 2025-04-15
Remove outdated code

eschrom:
All code specific to the autoregressive (i.e. Linear Incident Uptake Model) is removed. The Hill model is renamed to LL for "Logistic + Linear." Several other very minor changes.


eschrom:
This is now rebased on the latest changes. Although there appear to be a lot of changes here, they all serve the simple purpose of removing the autoregressive Linear Incident Uptake Model from the codebase.

docs/model_journal.md
```diff
@@ -30,10 +30,10 @@ Even with MCMC proceeding, other warning signs arose:
...
 
-## Solution Pt 2: Hill + Linear Functions
+## Solution Pt 2: Logistic + Linear Functions
 
-Many warning signs were alleviated by changing the structure of the latent true uptake from a pure Hill function to a Hill function plus a slope-only linear function (intercept = 0). In this model, the linear slope "M" and the Hill asymptote "A" can deviate additively from their overall averages by group, while the Hill midpoint "H" and steepness "n" are fixed across groups. In particular, this mixed function allows uptake to continue creeping upward late in a season.
+Many warning signs were alleviated by changing the structure of the latent true uptake from a pure Hill function to a logistic function (very similar to a Hill, but with a more computationally-friendly formulation) plus a slope-only linear function (intercept = 0). In this model, the linear slope "M" and the logistic asymptote "A" can deviate additively from their overall averages by group, while the logistic midpoint "H" and steepness "n" are fixed across groups. In particular, this mixed function allows uptake to continue creeping upward late in a season.
```
swo, line 35: I don't think you need to mention "Hill" here. Everyone knows logistic; only biologists know Hill, and I don't think it would be obvious how that applies to vaccine uptake

iup/models.py
```diff
@@ -947,7 +389,7 @@ def predict(
             season_start_day,
         )
 
-        scaffold = HillModel.augment_scaffold(
+        scaffold = LLModel.augment_scaffold(
```
swo, line 392: This is so minor but both "LL" (which sounds like log likelihood) and "logistic-linear" (which sounds like a GLM) seem funky

Maybe LPL for logistic plus linear? I was going to say LLM for logistic-linear mixture, but that's not good either

swo (approved):
Looks reasonable to me. A few suggestions. But I'm not super-close to this codebase, so it might be better to wait for more eyes.

docs/model_journal.md
eschrom: Haha coming from a bio background, I would have said nearly the opposite - "everyone knows Hill; only mathy people know logistic." But I'm happy to omit.

iup/models.py
eschrom: LPL works for me - happy to change it.

PR #165, opened 2025-04-15
Add test for parse_name_and_date()

Fuhan-Yang:
- Added unit test for `iup.utils.parse_name_and_date()`


swo (approved):
This code seems perfectly reasonable, but I'm still confused, as per #166

Issue #160, opened 2025-04-09
Arviz warning: Shape validation failed: input_shape: (1, 1000), minimum_shape: (chains=2, draws=4)


Fuhan-Yang:
Warning shows up when using this config: 

```
diagnostics:
  model: [HillModel]
  forecast_date: [2024-02-03]
  plot: []
  table: [print_model_summary]

```
The shape of `print_model_summary` is checked and correct. Maybe it's the Xarray output from arviz reporting the warning?

eschrom:
Addressed as of #167

Issue #120, opened 2025-02-12
Parallelize numpyro forecasts


swo:
Not sure if this is easy or worth it
```
python scripts/forecast.py --input=data/nis_raw.parquet --config=scripts/config.yaml --output=data/forecasts.parquet
WARNING:2025-02-12 15:40:38,297:jax._src.xla_bridge:987: An NVIDIA GPU may be present on this machine, but a CUDA-enabled jaxlib is not installed. Falling back to cpu.
/home/ulp7/proj/cfa-immunization-uptake-projection/iup/models.py:220: UserWarning: There are not enough devices
 to run parallel chains: expected 4 but got 1. Chains will be drawn sequentially. If you are running MCMC in CPU, consider 
using `numpyro.set_host_device_count(4)` at the beginning of your program. You can double-check how many devices are available in your system using `jax.local_device_count()`.
```

afmagee42:
It's both.* NumPyro is trying to say that all you need is one line before `run_all_forecasts` gets called in `main` in script/forecast.py:

```python
numpyro.set_host_device_count(4)
```

*The gains will be bigger the more intensive the model is to run. If you've got 4 CPUs available, and you run 1 chain on each, you should more or less get the same work done in 25% of the time it would take to run them sequentially. We do it in [linmod](https://github.com/CDCgov/cfa-viral-lineage-model/blob/1b3217bfb54e561f72b996a2ab65cc676822b4b2/retrospective-forecasting/main.py#L20) where models take 15+ minutes to run.

eschrom:
Addressed as of #167

Issue #110, opened 2025-01-28
Separate fitting and forecasting seeds


swo:
I think this is worth a separate discussion-- I think we might want separate fitting and prediction keys

Like, I think it's fair for us to fit the model a small number of times (1? 3?) but then we'll want to draw a number of trajectories (100? 1000?).

I want the same fitted model to produce the same trajectory when given the same input key.

There are some nice ways to generate child keys from a parent key.

_Originally posted by @swo in https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/102#discussion_r1932698564_

swo:
cc @afmagee42 if you have any deep thoughts here

afmagee42:
To the extent that prediction requires drawing new values, then from first principles I think having it be key-configurable is a good idea?

But I think we're in a danger zone for confusion here and need to be careful about what we mean and what's being separated from what. There are two distinct data-conditioned sets of things which we can refer to with prediction:
1. Draw observations from the observation model, this is the [posterior predictive distribution](https://en.wikipedia.org/wiki/Posterior_predictive_distribution). NumPyro provides this with [`Predictive`](https://num.pyro.ai/en/stable/utilities.html#numpyro.infer.util.Predictive) (which also includes the prior predictive, but that's not data-conditioned).
2. _Forecasting_ predictions; asking about what will happen in the future.

These aren't always neatly factorizable.

As of #125 with the Hill model, we have a model where they can be factorized. That model has a deterministic uptake curve and a model for how reality gets observed from it. For the Hill model, predictions about the latent true uptake in the future, that is (2) are deterministic given posterior samples, but that needn't be true in general. And for any time period, future or not, given the latent uptake (1) involves a "new" RNG call, in that if all you have are posterior samples, you can't just get this for free.

The autoregressive model, as currently implemented and as I understand it, hides this distinction by saying that what we've seen _is_ the uptake, so for the future (1) + (2) are the same, and in the past they're inapplicable.

afmagee42:
I suggest that the general easy mental model for thinking about this all is that the number of posterior predictive samples which is _natural_ for $n$ samples from the posterior distribution is
1. $k n$, where we usually have $k = 1$ and draw one new observation per posterior sample. This is what stan does. (We can also "thin" if we need to here, and have $k < 1$ and take fewer samples, or take $k > 1$ and draw more.)
[...]

PR #167, opened 2025-04-17
Parallelized numpyro chains

eschrom:
If there are multiple MCMC chains for model fitting (and there should be), they now run in parallel. MCMC control parameters are now separated from model parameters in the config to set up the parallelization a little more easily.

This resolves issue #120. 

This also resolves issue #160, because that Arviz warning was basically saying that we should run more than 1 MCMC chain.

This also resolves issue #110, because I also split the random seed that initializes the model into two children: one for model fitting and one for prediction.

### DIFF
--- a/iup/models.py
+++ b/iup/models.py
@@ -89,9 +89,10 @@ class LPLModel(UptakeModel):
 
         Parameters
         seed: int
-            The random seed for stochastic elements of the model.
+            The random seed for stochastic elements of the model, to be split for fitting vs. predicting.
         """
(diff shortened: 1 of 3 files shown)


swo (approved):
Looks good to me!

Is there some way to query the number of cores you have? I don't know if software typically warns you when you're doing more parallel jobs than cores. Out of scope for this PR but just curious

eschrom:
There must be a way to query your number of available cores - I've done it in R. We're safe with 4 here, but I'll look into it.

PR #172, opened 2025-04-25
Refactor plotting individual projections, summaries, and metrics

Fuhan-Yang:
- Updated `plot_scores()` to plot the metrics along forecast start. Even though the metrics between the data and the 2.5% and 97.5% percentile of forecasts are calculated, only shows the metrics between the data and the median of forecasts.
- Added group to plot in `plot_scores()`,  `plot_summary()` and `plot_inidividual_projections()`. So far, only allow to plot one group factor, as it will be the column dimension besides the existing row dimension such as `forecast_start` or `metric type` in the 2D plot. 
- Added an option in `config_template.yaml` to allow to choose which group factor to plot. 

The plot showing the individual projections of forecasts: 
image

The plot showing the mean and 95% prediction interval of forecasts:
image

The plot showing the evaluation metric changes over the forecast start date: 
image

### DIFF
--- a/scripts/config_template.yaml
+++ b/scripts/config_template.yaml
@@ -45,7 +45,12 @@ models:
       num_chains: 1
 
 # score metrics
-score_funs: [mspe, mean_bias, eos_abe]
+# score metrics
+scores:
+  quantiles: [0.025,0.5,0.975] # Must specify at least one
+  difference_by_date: [2024-03-31] # if blank, abs_diff will not be used
+  others: mspe
+  group_to_plot: [season]
 
 # Spacing between multiple forecasts, for evaluation
 evaluation_timeframe:
@@ -53,9 +58,9 @@ evaluation_timeframe:
 
 # Prediction credible interval plot
 forecast_plots:
-  interval: # need to be a percentile #
-    lower: 2.5
-    upper: 97.5
+  interval: # need to be a fraction #
+    lower: 0.025
+    upper: 0.975
   n_trajectories: 20
 
 diagnostics:
--- a/scripts/config_template.yaml
+++ b/scripts/config_template.yaml
@@ -47,9 +47,9 @@ models:
 # score metrics
 # score metrics
 scores:
-  quantiles: [0.025,0.5,0.975] # Must specify at least one
+  quantiles: [0.025,0.5,0.975] # Must specify at least one fraction
   difference_by_date: [2024-03-31] # if blank, abs_diff will not be used
-  others: mspe
+  others: [mspe]
   group_to_plot: [season]
 
 # Spacing between multiple forecasts, for evaluation
--- a/scripts/postprocess.py
+++ b/scripts/postprocess.py
@@ -84,7 +84,12 @@ def plot_individual_projections(
 
 
 def plot_summary(
-    obs: pl.DataFrame, pred: pl.DataFrame, groups: List[str,], lci: float, uci: float
+    obs: pl.DataFrame,
+    pred: pl.DataFrame,
+    groups: List[str,],
+    lci: float,
+    uci: float,
+    group_to_plot: List[str,],
 ):
     """
     Save a multiple-grid graph of observed data and mean, interval estimate of prediction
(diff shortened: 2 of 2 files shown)


scripts/postprocess.py
```diff
@@ -78,13 +83,24 @@ def plot_individual_projections(
...
 
+    if group_to_plot is not None:
+        assert len(group_to_plot) == 1
+        ("Only one grouping factor is allowed for score plot.")
+        group = group_to_plot[0]
+
     return alt.layer(pred_chart, obs_chart, data=pred).facet(
-        column="model", row="forecast_start"
+        row=alt.Row("forecast_start:T", title="Forecast start date"),
+        column=alt.Column(group, type="nominal"),
```
eschrom, line 93: In continuation of my above comment, this will lead to incorrect plots. Suppose your model groups by season, age, and state. Line 56 makes sure that only the most recent season is plotted, but that still leaves two grouping factors: age and state. If you choose age as the one and only grouping factor for plotting, then each column of plots represents an age group. But each age group still encompasses separate data and predictions for 50 states! So do you get 50 data series, plus 50 model trajectories, all crammed into each panel?

I think a better solution is to have model and forecast date be the rows and columns for faceting, and then make separate pages of these plots for each combination of grouping factors. For example, page 1 would be "Age 18-49 in Alabama," page 2 would be "Age 50-69 in Alabama," page 3 would be "Age 70+ in Alabama," page 4 would be "Age 18-49 in Alaska," and so on.

Alternatively, you could give the user control, by making config entries for the rows and columns of the plots. We know we must separate plots by: model type, forecast date, and each grouping factor except season. If the user specifies which two of these factors to use as rows and columns, we know that each combination of all the others should be a separate page.

scripts/postprocess.py
```diff
@@ -109,6 +131,8 @@ def plot_summary(
     if "time_end" not in obs.columns or "estimate" not in obs.columns:
         ValueError("'time_end' or 'estimate' is missing from obs.")
 
+    assert len(pred["model"].unique()) == 1, "Only 1 model is allowed. "
```
eschrom, line 134: Although the codebase only features one model right now, we should certainly allow multiple models here. If that creates too many categories to facet over, we should just use the strategy I suggested above to get multiple pages of faceted plots.

If there is more than one model, the entire pipeline will work up to this point, simply producing a .parquet of predictions spanning multiple models. But then this will suddenly break.

scripts/postprocess.py
```diff
@@ -161,42 +186,64 @@ def plot_summary(
...
 
+    if group_to_plot is not None:
+        assert len(group_to_plot) == 1
+        ("Only one grouping factor is allowed for score plot.")
+        group = group_to_plot[0]
+
     return alt.layer(interval_chart, pred_chart, obs_chart, data=plot_pred).facet(
-        column="model", row="forecast_start"
+        column=alt.Column(group, type="nominal"),
+        row=alt.Row("forecast_start:T", title="Forecast start date"),
```
eschrom, line 196: All the comments I have made above apply here too.

scripts/postprocess.py
```diff
@@ -215,12 +262,17 @@ def plot_score(scores: pl.DataFrame):
 
     pred = pl.read_parquet(args.pred)
     data = pl.read_parquet(args.obs)
+    scores = pl.read_parquet(args.score)
 
     with open(args.config, "r") as f:
         config = yaml.safe_load(f)
 
+    pred = pred.filter(pl.col("model") == pl.lit("HillModel"))
```
eschrom, line 270: As mentioned above, I don't think we should be restricting the plotting capabilities to one model only. We (or others) may well add new models soon, and the grouping factor issue forces us to think about faceting vs. paging now, so we might as well incorporate multiple models into that solution. (Also, "HillModel" is now replaced with "LPLModel.")

eschrom (changes requested):
My comments center around just one issue: we cannot restrict plotting to only one grouping factor when there are multiple (excluding season), because this will leave many separate data and prediction series (or evaluation scores) jumbled within a single panel, with no way to distinguish them. If this doesn't make sense, let me know! I'd be happy to talk through it.

I've suggested a strategy of using both faceting and multiple pages as a flexible solution, which I'm also happy to talk through.

Finally, I wanted to mention that I think your aesthetic choices for the plots are nice! Your examples look sharp!

scripts/config_template.yaml
```diff
@@ -46,22 +46,23 @@ mcmc:
...
   num_chains: 4
 
-# score metrics
 # score metrics
 scores:
-  quantiles: [0.025,0.5,0.975] # Must specify at least one
+  quantiles: [0.025,0.5,0.975] # Must specify at least one fraction
   difference_by_date: [2024-03-31] # if blank, abs_diff will not be used
   others: [mspe]
+  group_to_plot: [season]
```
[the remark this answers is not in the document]
Fuhan-Yang: Thanks for the suggestions! What grouping factors do we want to have for LPL model? Are there going to be "state", "age" and "season" in training, and "state" and "age" in prediction? I chose "season" because this is what we have in `config["data"]["groups"]`. Happy to talk synchronously!

Issue #171, opened 2025-04-22
Inconsistent appearance of divergences


eschrom:
When fitting the model on flu vaccination data, grouped by season and state, some random seeds produce chains with divergences, high R-hats, and low ESS. Investigate what is going wrong in these chains, and how to fix it. Tweaking priors and running longer warmups are my best guesses right now.

eschrom:
Fixed in #173

PR #173, opened 2025-04-29
Adjusted Priors in Small but Crucial Ways

eschrom:
I adjusted the suggested priors in the config template to give better MCMC behavior. The crucial change, I believe, is increasing the gamma prior for M (slope of the linear function) so that the prior probability approaches 0 as values of M approach 0. That is, while M is likely small, we are confident that it is also greater than 0.

Before making this change, 16 independent HMC chains across 4 different random seeds ranged in run times from 14 to 78 minutes, and the longer chains produced hundreds of divergences on rare occasions.

After this change, 20 independent HMC chains across 5 different random seeds ranged in run times from 10 to 13 minutes and produced at most 2 (but usually 0) divergences.

This is a tiny amount of code, but a big improvement. This addresses #171

### DIFF
--- a/scripts/config_template.yaml
+++ b/scripts/config_template.yaml
@@ -29,16 +29,16 @@ models:
     params:
       A_shape1: 100.0
       A_shape2: 180.0
-      A_sig: 40.0
+      A_sig: 80.0
       H_shape1: 100.0
       H_shape2: 225.0
-      n_shape: 25.0
-      n_rate: 1.0
-      M_shape: 1.0
-      M_rate: 10.0
-      M_sig: 40
-      d_shape: 350.0
-      d_rate: 1.0
+      n_shape: 150.0
+      n_rate: 6.0
+      M_shape: 2.0
+      M_rate: 25.0
+      M_sig: 80.0
+      d_shape: 1000
+      d_rate: 3.0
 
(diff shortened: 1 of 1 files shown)


eschrom:
Troubleshooting MCMC is a mix of trial-and-error and structured logic. The trace plot diagnostics you wrote were a _fantastic_ help here. 

Essentially, I noticed that the troublesome chains estimated the slope of the linear function ($M$) to be slightly smaller, and the magnitude of state-to-state variation ($\sigma_A$ and $\sigma_M$) to be much larger, compared to the healthy chains. So I first shrunk the priors for $\sigma_A$ and $\sigma_M$ to try to keep the chains from exploring values for these that were too high. But surprisingly, that didn't work. 

Then I suspected that the chains might be forced to defying the priors with really high values of $\sigma_A$ and $\sigma_M$ to compensate for too-low values of $M$. So I adjusted the prior for $M$ to better reflect our legitimate belief that $M$ is small but not 0. In particular, I changed the prior for $M$ from: 
image
to:
image

With this change, the chains all began behaving consistently well. So there was some trial-and-error and some logic behind realizing that $M \rightarrow 0$ was leading the chains into a problematic corner of parameter space, and using priors to prevent that.

PR #175, opened 2025-05-01
Excise parse_name_and_date

### LINKED ISSUE #166: Don't serialize model fit IDs to a string?
I'm looking at #165. I had a comment to this effect elsewhere too. Rather than taking the model name (a string) and a date, turning that into a string to use a dictionary key, and then parsing that string-- why not just keep the tuple?

```python
{
  ('model_name', datetime.date(2025, 1, 1)): 'my_first_fit_here',
  ('model_name', datetime.date(2025, 1, 8)): 'my_second_fit_here'
}
```

Fuhan-Yang:
Per discussion in #166, it's more clear to represent the keys in model dictionary using tuple than string. Excise `parse_name_and_date()`.

Resolves #166

### DIFF
--- a/scripts/diagnostics.py
+++ b/scripts/diagnostics.py
@@ -9,7 +9,6 @@ import yaml
 import iup
 import iup.diagnostics
 import iup.models
-from iup.utils import parse_name_and_date
 
 
 def diagnostic_plot(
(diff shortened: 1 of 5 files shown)


scripts/diagnostics.py
```diff
@@ -29,7 +28,9 @@ def diagnostic_plot(
             plot_func = getattr(iup.diagnostics, plot_name)
             axes = plot_func(idata)
             fig = axes.ravel()[0].figure
-            fig.savefig(f"{output_dir}/{key}_{plot_name}.png")
+            fig.savefig(
+                f"{output_dir}/{key[0]}_forecast_start_{str(key[1])}_{plot_name}.png"
```
eschrom, line 32: We can probably omit "forecast_start" since any date that appears in the filename of a figure can be assumed to be the forecast start date. Your call, though!

eschrom (approved):
Looks great, Fuhan!!

scripts/diagnostics.py
Fuhan-Yang: I will keep it for now, and can remove later if it should!

PR #176, opened 2025-05-01
Miscellaneous refactor on diagnostics

### LINKED ISSUE #159: Can't change the default value of max number of subplots in Arviz
```
az.rcparams.defaultParams["plot.max_subplots"] = 100
```
Still get warning: `UserWarning: rcParams['plot.max_subplots'] (40) is smaller than the number of resulting pair plots with these variables, generating only a 8x8 grid`

Fuhan-Yang:
- Only plot the parameters without "*_devs" in `iup.diagnostics.parameter_pairwise_plot()`.
- Update function signature from `idata` to `UptakeModel` to enable more control in the functions.
- Remove $(POSTERIOR) target in Makefile. Now we can use `print_posterior_dist()` in `iup.diagnostics` to save it!

Resolves #159

### DIFF
--- a/iup/diagnostics.py
+++ b/iup/diagnostics.py
@@ -4,62 +4,71 @@ import polars as pl
 import iup.models
 
 
-def posterior_density_plot(idata: az.InferenceData) -> dict:
+def posterior_density_plot(model: iup.models.UptakeModel) -> dict:
+    idata = az.from_numpyro(model.mcmc)
     return az.plot_posterior(idata)
 
 
-def parameter_trace_plot(idata: az.InferenceData) -> dict:
+def parameter_trace_plot(model: iup.models.UptakeModel) -> dict:
+    idata = az.from_numpyro(model.mcmc)
     return az.plot_trace(idata)
 
 
-def parameter_pairwise_plot(idata: az.InferenceData) -> dict:
-    return az.plot_pair(idata)
+def parameter_pairwise_plot(model: iup.models.UptakeModel) -> dict:
+    idata = az.from_numpyro(model.mcmc)
+    if isinstance(model, iup.models.LPLModel):
+        # remove A_devs and M_devs for LPL model
+        return az.plot_pair(idata, var_names=["~A_devs", "~M_devs"])
+    else:
+        return az.plot_pair(idata)
 
 
-def print_posterior_dist(
-    fit_model: iup.models.UptakeModel, idata: az.InferenceData
-) -> pl.DataFrame:
+def print_posterior_dist(model: iup.models.UptakeModel) -> pl.DataFrame:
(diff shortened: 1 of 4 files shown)


iup/diagnostics.py
```diff
@@ -4,62 +4,71 @@
...
 
 
-def print_posterior_dist(
-    fit_model: iup.models.UptakeModel, idata: az.InferenceData
-) -> pl.DataFrame:
+def print_posterior_dist(model: iup.models.UptakeModel) -> pl.DataFrame:
+    idata = az.from_numpyro(model.mcmc)
     posterior = idata.to_dataframe(groups="posterior", include_coords=False)
     posterior = pl.from_pandas(posterior)
+
```
eschrom, line 30: In case you hadn't seen my comment elsewhere, I check in with another ArviZ aficionado who confirmed that routing through Pandas is indeed the best way to do this.

iup/models.py
```diff
@@ -76,6 +76,9 @@ def predict(
         """
         pass
 
+    # save for future models #
+    mcmc = None
```
eschrom, line 80: Could you explain what this is? My best guess is that the model must have an `mcmc` attribute in order for the code in `iup/diagnostics.py` to work. But I would have guessed this would say `self.mcmc = None` rather than `mcmc = None`. What am I misunderstanding?

scripts/diagnostics.py
```diff
@@ -43,14 +40,11 @@ def diagnostic_table(
     diagnose_table_names = config["diagnostics"]["table"]
 
     for key, model in sel_model_dict.items():
-        idata = az.from_numpyro(model.mcmc)
+        # idata = az.from_numpyro(model.mcmc)
```
eschrom, line 43: Scott always tells me to remove code altogether rather than commenting it out. Just passing this along haha

eschrom (approved):
This also looks great!

iup/diagnostics.py
Fuhan-Yang: Good to know that we are on the right track!

iup/models.py
Fuhan-Yang: Your guess is right! I think we will use numpyro for future models so added this placeholder. Not using `self.mcmc = None` is because, `self` is not defined here, as we are not in a function with argument like `f(self, arg1, arg2)`, thus python doesn't recognize what `self` is. By just putting `mcmc = None` under class `UptakeModel`, `mcmc` is directly added as an attribute, and then it can be indexed using `iup.UptakeModel.mcmc`.

scripts/diagnostics.py
Fuhan-Yang: Oops! I must have skipped it. Thanks for flagging!!

PR #178, opened 2025-05-02
Restructure output files

eschrom:
Output files from the pipeline are restructured so that only one file type appears in each output folder. Here is the output scheme:
image
This gives the directory structure as well as the file type inside each folder.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -28,7 +28,7 @@ $(SCORES): scripts/eval.py $(FORECASTS) $(RAW_DATA)
 $(FORECASTS): scripts/forecast.py $(RAW_DATA) $(MODEL_FITS) $(CONFIG)
 	python $< --input=$(RAW_DATA) --models=$(MODEL_FITS) --config=$(CONFIG) --output=$@
 
-$(DIAGNOSTICS) $(DIAGNOSTIC_PLOTS): scripts/diagnostics.py $(MODEL_FITS) $(CONFIG)
+$(DIAGNOSTICS): scripts/diagnostics.py $(MODEL_FITS) $(CONFIG)
 	python $< --input=$(MODEL_FITS) --config=$(CONFIG) --output_table=$(DIAGNOSTICS) \
 	--output_plot=$(DIAGNOSTIC_PLOTS)
 
--- a/Makefile
+++ b/Makefile
@@ -1,24 +1,23 @@
 NIS_CACHE = .cache/nisapi
 TOKEN_PATH = scripts/socrata_app_token.txt
 TOKEN = $(shell cat $(TOKEN_PATH))
-CONFIG = scripts/config.yaml
-RAW_DATA = data/nis_raw.parquet
-MODEL_FITS = output/model_fits.pkl
-DIAGNOSTICS = output/
-FORECASTS = data/forecasts.parquet
-SCORES = data/scores.parquet
-PROJ_PLOTS = output/projections.png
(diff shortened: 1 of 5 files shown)


scripts/diagnostics.py
```diff
@@ -26,7 +26,7 @@ def diagnostic_plot(
             axes = plot_func(model)
             fig = axes.ravel()[0].figure
             fig.savefig(
-                f"{output_dir}/{key[0]}_forecast_start_{str(key[1])}_{plot_name}.png"
+                f"{output_dir}{key[0]}_forecast_start_{str(key[1])}_{plot_name}.png"
```
swo, line 29: I try to avoid constructing paths by hand for exactly this reason, that you need to be careful about the slashes

I recommend `from pathlib import Path; Path(output_dir, f"whatever")`

swo: And as a sidebar, the Hive formatting is `key=value`, which helps clarify long filenames for me. Eg this could have `"forecast_start=2025-01-01"` in it

scripts/postprocess.py
```diff
@@ -221,12 +220,16 @@ def plot_score(scores: pl.DataFrame):
 
     plot_individual_projections(
         data, pred, config["forecast_plots"]["n_trajectories"]
-    ).save(args.proj_output)
+    ).save(f"{args.output}trajectories.png")
```
swo, line 223: see above, I think `Path()` is your friend here

swo (approved):
Some minor suggestions

It's not crazy for some of the outputs, if they are small, to be .csv rather than .parquet

scripts/postprocess.py
eschrom: Sounds great - I will learn how to do this.

scripts/diagnostics.py
swo: The Hive format is to have nested folders:

```
forecast_date=2025-01-01/
  state=AL/
    age_group=18-65/
      part-0.parquet
```

My point here is just that `forecast_start=2025-01-01.parquet` is easier for me to read than `forecast_start_2025-01-01`; it makes it clear that 2025-01-01 is a forecast start date. But Hive only works with things like `polars.scan_parquet()` or `scan_csv`; the only benefit here is human readability (and filename parsing, if you ever need to do that)

eschrom:
`Path` from `pathlib` is now incorporated, so is the `key=value` motif in filenames where appropriate (for readability rather than true hive structure - see above).

Beware when running `make` that `postprocess.py` currently breaks; this will be addressed in #172

Makefile
```diff
@@ -1,35 +1,36 @@
...
-CONFIG = scripts/config.yaml
-RAW_DATA = data/nis_raw.parquet
-MODEL_FITS = output/model_fits.pkl
-DIAGNOSTICS = output/
-FORECASTS = data/forecasts.parquet
-SCORES = data/scores.parquet
-PROJ_PLOTS = output/projections.png
-SUMMARY_PLOTS = output/summary.png
-SCORE_PLOTS = output/scores.png
+CONFIG = scripts/config_flu.yaml
```
Fuhan-Yang, line 4: Why do we change to `config_flu` here?

Fuhan-Yang (commented):
Thanks for separating the plots and parquet into different folders, the project looks more clear now!

Makefile
eschrom: That's just a separate config file I have that is designed to work with the flu data.

PR #179, opened 2025-05-02
Update Makefile workflow for new users

swo:
- The `.PHONY` thing meant that I had to specifically call `make cache` rather than just `make`
- Without a placeholder, I had no `data/` folder and was getting some error


eschrom (approved):
Looks good. I had to add a bunch of placeholders in #178 for the same reason.

Fuhan-Yang (approved):
Looks good to me!

Issue #174, opened 2025-05-01
Multiple pages of plots using Altair


Fuhan-Yang:
There are multiple facets we need to include when plotting the forecasts from LPL model: forecast start, model, grouping factors from data: at least age, state and season. As Ed suggested [here](https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/172#discussion_r2068538641), we can have the plot faceted by model and forecast start date on one page, and each page represent a combination of levels in grouping factors (Like "Age 18+, Alabama", "Age 0-17, New York"). There are two ways to generate the multiple pages of plot:

1. Matplotlib has backend "PdfPages" can write all the plots into a multiple-page single PDF. We need to convert altair object to matplotlib object, and save as one page using "PdfPages".
2. Save each plot separately and name it as the levels in the grouping factor combo.

For the current model, we can also neglect "season" and "model" as we only have one model and we only predict the most recent season. Then the plot will be faceted by forecast start date and other grouping factors, where we still need multiple pages to include all grouping factors. 

What do you think? @swo @eschrom

swo:
It makes sense to me to have 1 file per 2 dimensions, eg states on rows, age groups on columns, model = color. And then if you need another dimension (eg score?) put that in a different file.

I don't see a particular need to have nicely formatted pdf's

eschrom:
Forecast date would be a likely 4th+ dimension. 

I hadn't considered color, but that would work well for dimensions that only have small number of values (like you said, model). 

We might want to invest a little thought in the best way to choose the three on-page dimensions vs. the other across-page dimensions. E.g. states on rows could be problematic because there are >50 states, which might make the plots tiny/obscured?

Some solutions might be:
- Make the user choose what the columns/rows/colors/pages are in the config. (But there is already a lot in the config...)
- Choose the three dimensions with the most values, but not exceeding some upper limit (rows=12, cols=12, colors=5?), to be on-page, and make the rest across-page.

Are there are better solutions that I'm not seeing? I can always imagine a plausible scenario where the plotting gets ugly. Maybe I'm just being too anxious.

swo:
Free your mind from the 8.5x11 page! The png can be as big as you want, and you can scroll around it

Also, maybe this is a sign that you don't actually want to produce files of plots. Why not build a simple streamlit interface that lets you pick which age group, forecast date, etc. you want to look at?

Fuhan-Yang:
[...]

PR #180, opened 2025-05-02
Demo streamlit app

swo:
See #174 

You'll need to `poetry install` to get the new dependencies and then `make viz` to run the app.

This is just a starting point; I imagine that the plots you'll want to make are evolving. But at least this show how it can be done.

image


Fuhan-Yang (commented):
It turns out it's not that hard to develop an app! It's great to see that streamlit can directly embed altair object in the app. I'm happy to try streatmlit, as I think it is informative to have such app in our final product. But that also means we will take more time to finish the project. I can add as one ticket in the sprint. 

It seems hard to have all the subplots by the combination of group factors in one page. We can either visualize a single plot that is a certain combo by selecting the levels for each group factors, or visualize every level in a group factors given it is from a certain level of another group factor.

PR #182, opened 2025-05-13
Posterior Checks

eschrom:
While making plots for a demo presentation, I realized:
1) We generate forecasts to compare vs. the test data, but no posterior checks to compare vs. the training data.
2) The model's forecasts are substantially worse than the model's fit to training data.
3) The pipeline at times feels clunky, and it can be tricky to not overwrite results that took a long time to get.

This PR only deals with 1. I will make issues for 2 (#185) and 3 (#186).

Essentially, I added posterior checks as an output of the `forecast.py` script, and I adjusted the Makefile accordingly. 

The other altered file, `figs_demo.py`, is my notebook-in-progress to make custom figures for the demo presentation that sparked all this. No need to review this in great detail, since we may end up pruning it out of the repo later. But if you want to take a look and comment on whether we should make lasting use of this (e.g. as a demo that lives in the repo), feel free.

### DIFF
--- a/scripts/figs_demo.py
+++ b/scripts/figs_demo.py
@@ -1,0 +1,148 @@
+# %% Imports
+import altair as alt
+import polars as pl
+
+
+# %% Define some functions
+def load_data(path):
+    return (
+        pl.read_parquet(path)
+        .with_columns(
+            start=(pl.col("season").str.slice(0, 4) + pl.lit("-07-01")).str.strptime(
+                pl.Date, "%Y-%m-%d"
+            ),
+            obs_upper=pl.col("estimate") + 1.96 * pl.col("sem"),
+            obs_lower=pl.col("estimate") - 1.96 * pl.col("sem"),
+        )
+        .with_columns(elapsed=(pl.col("time_end") - pl.col("start")).dt.total_days())
+        .drop(["start", "sem", "N_vax", "N_tot"])
+        .rename({"estimate": "obs"})
+    )
+
+
+def load_pred(path, data):
+    pred = pl.read_parquet(path).drop(["forecast_start", "forecast_end", "model"])
+    return (
+        pred.group_by(["time_end", "geography", "season"])
+        .agg(
+            est=pl.col("estimate").mean(),
+            est_upper=pl.col("estimate").quantile(0.975),
+            est_lower=pl.col("estimate").quantile(0.025),
+        )
+        .join(
+            data,
+            on=["time_end", "geography", "season"],
+            how="left",
+        )
+    )
+
+
+def plot_uptake(data, color="green"):
+    plot_list = []
+    if "season" in data.columns:
+        plot_list.append(
+            alt.Chart(data)
+            .mark_line()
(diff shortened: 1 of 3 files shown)


scripts/figs_demo.py
```diff
@@ -0,0 +1,148 @@
...
+plot_uptake(flu_natl)
+
+# %% Plot state scale flu vax across seasons
+alt.data_transformers.disable_max_rows()
+plot_uptake(flu_state)
+
+# %% Plot uptake for one state in one season, with empirical uncertainty
+flu_state_sub = flu_state.filter(
+    (pl.col("geography") == "California") & (pl.col("season") == "2023/2024")
+).drop(["geography", "season"])
```
Fuhan-Yang, line 118: Curious about why plotting California/2023-2024, in additional to Missouri and Wyoming in #185.

scripts/forecast.py
```diff
@@ -47,12 +48,33 @@ def run_all_forecasts(
...
             test_dates = test_data.select(["time_end", "season"])
 
+        postcheck = fitted_model.predict(
+            start_date=train_data["time_end"].min(),
+            end_date=train_data["time_end"].max(),
+            interval=config["forecast_timeframe"]["interval"],
+            test_dates=train_dates,
+            groups=config["data"]["groups"],
+            season_start_month=config["data"]["season_start_month"],
+            season_start_day=config["data"]["season_start_day"],
```
Fuhan-Yang, line 67: This step can be made available in `diagnostics.py`. In that file, each fitted model will be called, with their unique forecast start to get their own training data (from `split_train_test`). This makes it easy to transfer posterior check in `diagnostics.py`.

Fuhan-Yang (commented):
This generally looks good to me! My general comment is that posterior check is a diagnostic tool of model fitting, and should be included in `scripts/diagnostics.py`, not `forecast.py`. If we decide to keep this functionality for future use, I think we should categorize it into `scripts/diagnostics.py` and make related refactoring. But if we only want to have a quick diagnostics and may remove it for the final product, I'm happy to keep as it is

scripts/figs_demo.py
eschrom: This is just an arbitrary choice of state and season. This notebook is meant to be interactive, where you can plug in whatever state and season choices you want.

scripts/forecast.py
eschrom: Conceptually, posterior checks are used as a diagnostic. But they are computed exactly the same way that forward predictions are computed. That is, once it's fit, a Bayesian model doesn't "know" whether it is doing posterior or forward prediction. This makes the code fit more efficiently into the `forecast.py` script. In fact, it might make sense to generate posterior and forward predictions (i.e. `postcheck` and `forecast` in `forecast.py` together in just one python object, and only split them later for scoring and plotting. This would be a minor refactor that I would be happy to do. As I work more with the posterior checks in the coming weeks, I'll get a better sense of how they could be handled most conveniently.

PR #191, opened 2025-05-16
Update README

Fuhan-Yang:
- Update package descriptions. 
- Update workflow diagram in par with the updated features in the package. It seems pretty nice! This can be useful for demo slides.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,21 +40,34 @@ Use <https://github.com/CDCgov/nis-py-api> for access to the NIS data.
 flowchart TB
 
 nis_data(nis_raw.parquet)
+fits(model_fits.pkl)
+diagnostic_table(diagnostic_tables.parquet)
+diagnostic_plot(diagnostic_plots.png)
 forecast(forecasts.parquet)
 scores(scores.parquet)
 config{{config.yaml}}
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -40,58 +38,78 @@ Use <https://github.com/CDCgov/nis-py-api> for access to the NIS data.
...
-subgraph score plots
-score_plot
+subgraph streamlit
+direction LR
+proj_plot~~~pred_summary~~~score_plot
 end
 
 NIS -->preprocess.py --> nis_data
+nis_data --> fit.py --> fits
+fits --> diagnostics.py
```
eschrom, line 83: I really like your colors and organizational scheme! Especially the boxes that group separate outputs from the same script. 

One edit: I think we need arrow from `model_fits.pkl` to `forecast.py`, maybe in place of the arrow from `nis_raw.parquet` to `forecast.py`.

eschrom (approved):
One adjustment to the flow diagram, but otherwise it looks great!

README.md
Fuhan-Yang: Good catch! Fixed!

PR #194, opened 2025-05-30
Fixed Bug in Forecasting

eschrom:
We previously noticed that although the model performed well on training data, it performed terribly when forecasting. I have traced this problem back to a bug in the model fitting/prediction code and fixed it. The changes to `iup/models.py` and `iup/utils.py` are the key changes.

Feel free to ignore the mass of sloppy code in `scripts/figs_demo.py` - this is merely the scratchwork I used to make plots as I investigated the aforementioned bug. I want to let this scratchwork hang around a few weeks, in case any of it proves repeatedly useful. But most of it will eventually disappear, so it is not worth your time to review.

I will circulate a more detailed explanation of my model diagnostic process in another channel, which will give the context for why I've made the changes I did here. Also happy to discuss it synchronously.

### DIFF
--- a/scripts/figs_demo.py
+++ b/scripts/figs_demo.py
@@ -136,7 +136,7 @@ postcheck_sub = postcheck.filter(
 plot_uptake(postcheck_sub)
 
 # %% Plot posterior predictions for all states in one season
-postcheck_sub = postcheck.filter(pl.col("season") == "2022/2023").drop("season")
+postcheck_sub = postcheck.filter(pl.col("season") == "2020/2021").drop("season")
 alt.data_transformers.disable_max_rows()
 plot_uptake(postcheck_sub)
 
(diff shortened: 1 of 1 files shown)


Fuhan-Yang (commented):
My understanding is: In hierarchical modeling, we sampled the deviation for each level in each group, then we summed up the deviations weighted by group-specific deviations across different groups. To do this we need unique ID for each level of each group. Previously we propagated the index from 0 to N from the first level of the first group, to the last level of the last group. But that comes from the data, which share different indices with the forecast. Now, we specified the indices from the forecast. 

My question is: how does adding the number of unique levels can make the model to recognize the indices from the forecast? This value will be assigned from training data during training, how does this value be assigned for the forecast?

Fuhan-Yang (approved):
That sounds a pretty smart trick! Happy to approve even I haven't totally understood. Thanks for the efforts, the model indeed seems have improved tremendously!

eschrom:
It certainly does make our forecasts look quite a bit better! I suspect our strategies for indexing in numpyro will evolve in the long term, but I'm happy with this. Thank you!

Issue #189, opened 2025-05-16
output/data/.placeholder


swo:
If you run `make` from a clean repo, you get an error that this folder doesn't exist

Issue #186, opened 2025-05-14
Smooth out the pipeline


eschrom:
As I diagnose and debug models, I often get stuck either running the pipeline piecemeal with single `python scripts/forecast.py ...` commands, or else overwriting files unintentionally with Make. To ease these annoyances, it would be useful to define a single nickname for my entire pipeline run at the top of the Makefile, and then include this nickname in all output files. Or something like that. I will keep observing the inconveniences I encounter and then design some solution.

eschrom:
This is sorted out with a revamp of the Makefile in #195

PR #195, opened 2025-06-05
Nicknames in the Makefile

eschrom:
Line 1 of the Makefile is now a nickname that is assigned to a particular model fitting + forecasting + scoring run. After you `make all`, if you `make all` again without changing the nickname, Make only runs steps for which a target and/or prerequisite has changed, as expected. But if you `make all` again with a different nickname, the entire pipeline runs again, and its output is stored without overwriting the previous run. This addresses #186. 

To accomplish this, when `make all` is run with a new nickname, 5 new subdirectories (named using the nickname) are created within the 5 output directories (data, fits, diagnostics, forecasts, and scores) to hold the corresponding outputs. Just a little extra work using `Path` from `pathlib` in each of the scripts enables the dynamic creation of these directories. Placeholder files are not required, and many are removed, so this addresses #189.

`make clean` will remove all 5 of these nicknamed subdirectories.

There are also several other minor changes for the sake of convenience:
- The `plot` vs. `table` distinction in the output directories is removed, reducing the number of targets that need to be defined in the Makefile and the number of arguments that several of the scripts need.
- The interface with the NIS API in the Makefile is simplified, so that `make nis` will cache all available NIS API data sets (warning and skipping any that already exist in the cache, so as to not overwrite), and `make delete_nis` will clear the cache of all NIS API data sets.
- The cache for the NIS API data sets is now the global system default (typically `/home/<userid>/.cache/nisapi`) to avoid the need for a local NIS data cache inside the repo.
- The Makefile's default config file `config_template.yaml` is changed to only perform one forecast, not multiple on different forecast dates. This is merely for faster testing of the pipeline. Multiple forecasts for evaluation purposes can be added back in at any time.
- Many of the packages have been updated, including updating the NIS API to the latest version, as reflected in `poetry.lock`.
- Some `CumulativeUptakeData` and `QuantileForecast` objects must be coerced into standard data frames in the `summarize_score` function of `iup/eval.py` to make their types match for joining. This wasn't a problem before, so I suspect that updating the package version of polars introduced this new requirement.

This is more for @Fuhan-Yang to review than @afmagee42, but I thought you both might be interested in the nicknaming strategy.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -38,6 +38,9 @@
 *.Identifier
 *.pyc
 
+# Config copies
+output/
+
 # Byte-compiled / optimized / DLL files
 __pycache__/
 *.py[cod]


Makefile
```diff
@@ -1,42 +1,46 @@
...
+clean:
+	rm -r $(RAW_DATA) $(MODEL_FITS) $(DIAGNOSTICS) $(PREDICTIONS) $(SCORES)
+
+nis:
+	python -c "import nisapi"
+	python -m nisapi cache --app-token=$(TOKEN)
+
+delete_nis:
+	python -c "import nisapi"
+	pyton -m nisapi delete
```
afmagee42, line 46:
```suggestion
	pytjon -m nisapi delete
```

scripts/diagnostics.py
```diff
@@ -94,16 +94,16 @@ def select_model_to_diagnose(models: Dict[str, iup.models.UptakeModel], config)
 if __name__ == "__main__":
     p = argparse.ArgumentParser()
     p.add_argument("--config", help="config file")
-    p.add_argument("--input", help="fitted models")
-    p.add_argument("--output_table", help="path of output tables")
-    p.add_argument("--output_plot", help="path of output plots")
+    p.add_argument("--input", help="fitted model directory")
+    p.add_argument("--output", help="output directory")
```
swo, line 98: My typical recommendation is to have the file organization rest at the highest level (eg the Makefile), rather than in individual scripts, and instead have scripts take file paths only. This way, if you ever want to change the file organization, you only need to go to one place

But naturally I break this rule all the time, especially for outputs. It's much easier to say "script, write these 4 files with these 4 names to this 1 directory" rather than have four command line arguments

Makefile
```diff
@@ -1,42 +1,46 @@
...
 TOKEN = $(shell cat $(TOKEN_PATH))
-CONFIG = scripts/config.yaml
-RAW_DATA = output/data/nis_raw_flu.parquet
-MODEL_FITS = output/fits/model_fits.pkl
-DIAGNOSTICS = output/diagnostics/tables/
-DIAGNOSTIC_PLOTS = output/diagnostics/plots/
-POSTCHECKS = output/forecasts/tables/postchecks.parquet
-FORECASTS = output/forecasts/tables/forecasts.parquet
-SCORES = output/scores/tables/scores.parquet
+CONFIG = scripts/config_template.yaml
```
swo, line 4: I'd recommend copying your config somewhere, so you know what config you were using in association with which model run

Makefile
```diff
@@ -1,42 +1,46 @@
-NIS_CACHE = .cache/nisapi
+NICKNAME = test_run
```
swo, line 1: I'd recommend calling this RUN_ID or something like that that's more consistent with what other teams/groups would call this

swo: I would call it just `RUN_ID=test`, or `dev`, or something, rather than have "run" in there too.

It might be worth making a note in the readme that you can [override variables](https://www.gnu.org/software/make/manual/html_node/Overriding.html) with make. So you could `make RUN_ID=my_next_run` rather than editing the Makefile.

Makefile
swo: Really the next level here is to use a random run ID (like a runtime concatenated with a short uuid), and then you'd basically be browsing through your configs for the run you want

I suppose that defeats the idea of "nicknames," but it does mean that you don't need to ever worry about overwriting stuff

swo (approved):
I have some comments, but I'm not here to stand in your way; I'll make sure you're approved in case you need that to go on

PR #200, opened 2025-07-15
Refactor streamlit

### LINKED ISSUE #187: Choose quantile of prediction interval in streamlit
I think these quantiles were config entries under `config["forecast_plots"]`, so those values should be substituted in here. That said, your idea to have more such entries be made directly in streamlit, rather than the config, is a very good one!

_Originally posted by @eschrom in https://github.com/CDCgov/cfa-immunization-uptake-projection/pull/183#discussion_r2091782470_

### LINKED ISSUE #190: Single data frame of observed data and forecasts in streamlit
Quote from #183:

"Currently, if users choose geography in one of the dimensions, all the observed data across states will show in each faceted plot, instead of the data only from that state. This is more an Altair problem. When there is a layered and faceted plot (multiple charts of data and prediction), it is supposed to only show the value under the corresponding group factors for both data frames. One solution is that to aggregate the data with the forecasts, and only deal with a single data frame. But that will lose the feature that observed data is showed as points, while forecasts are showed as lines, which can be important distinction in the individual trajectory plot."

Fuhan-Yang:
- Debugged altair by having one single data source for multiple charts. 
Now if `make viz`, we should see trajectories of observed data along with other sample predictions; prediction interval (95% for now, can change in yaml) with mean trajectories and points from observed data; point plot of MSPE and absolute error of end-of-season uptake. 
- Integrated to the new Makefile structure. Connect streamlit as proposed in #195.
- Working on adding interaction: when mouse hovers on the plot, data will show up. 

Problems to solve: stop responding when plotting `geography` dimension, may be related to cache

Resolves #187 #190

### DIFF
--- a/scripts/viz.py
+++ b/scripts/viz.py
@@ -16,9 +16,6 @@ def app():
     obs = data["observed"]
     pred = data["forecasts"]
 
-    states = ["Alabama", "California", "Virginia"]
-    obs = obs.filter(pl.col("geography").is_in(states))
-    pred = pred.filter(pl.col("geography").is_in(states))
     # load config
     config = load_config()
 
@@ -50,8 +47,6 @@ def plot_trajectories(obs: pl.DataFrame, pred: pl.DataFrame, config: Dict[str, A
         The configuration yaml file.
 
     """
-
-    obs = obs.with_columns(model=pl.lit("observed"))
 
     # set up plot encodings
     encodings = {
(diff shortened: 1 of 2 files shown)


scripts/viz.py
```diff
@@ -370,30 +399,34 @@ def layer_with_facets(data: pl.DataFrame, *charts: List, **encodings: Dict):
...
+        }
 
-@st.cache_data
-def load_config():
-    return yaml.safe_load(open("scripts/config.yaml"))
+    @st.cache_data
+    def load_scores():
+        return pl.read_parquet(Path(args.score, "scores.parquet"))
 
+    @st.cache_data
```
eschrom, line 428: Having not experimented with caching myself, I'm a little unsure of how this works. Caching the data, forecasts, and scores makes sense, because this stuff should never change within a streamlit session. But what if you want to change a config setting (e.g. number of trajectories or credible interval) to plot within a streamlit session? Does this mean you would have to launch a new session to recognize config changes? This is a very minor thing - just curious for my own understanding.

eschrom (approved):
These plots look fantastic, @Fuhan-Yang! While I have been reading a lot about streamlit, I think I've reached the point where I'll learn more by experience than by extra code reading. If further figure changes prove necessary, I can get some streamlit experience by addressing them incrementally and writing small PRs. But this looks very good to me!

scripts/viz.py
Fuhan-Yang: Yes, it needs a new session to update the changes in config (refreshing browser doesn't help). I guess it's because the config won't be updated when the session is called. To change something that can instantaneously reflect in the app, we should use [input widgets](https://docs.streamlit.io/develop/api-reference/widgets) in streamlit

PR #209, opened 2025-09-08
More GAM exploration

Fuhan-Yang:
More exploratory analysis of GAM. 

Mathematically, the GAM model we used is: 
$Y = X\beta+ Zu + \beta_0$

$Y$ is the observed vaccine uptake, $X$ is the design matrix of basis function with N * k dimension, where N is the number of observations and k is the number of basis functions used. Each element in $X$ is the value of the basis function evaluated at the predictor `elapsed`. $\beta$ is a vector of coefficients that control each basis function. $X\beta$ is the main effect that is the same across all the level in a group (in our case is all seasons in `season` ).

$Z$ is the covariance matrix of each grouping factor, and $u$ is group-specific intercept. $Z$ is an identity matrix that assumes each level has effect on $Y$ independent from other levels. $Zu$ is the random effect that models the level-specific intercept. 

$beta_0$ is the main intercept. 

When fitting the model with national flu vaccine uptake, we found that the fitted line by each season shares the same shape and the intercept differs by season. 
fitted_value_by_season

To confirm that, if we exclude the group-specific intercept, the main effect fitted by each season is the same
main_effect_by_season

When predicting, the predicted group-specific intercept is 0 for the unobserved group, resulting that only the main effect trained by observed seasons is showed: 
pred_by_sel_seasons
Here we respectively used 2, 4, 8, 12, and 14 seasons in training and showed the prediction. 

It is expected to see that MSPE decreased as more seasons are included in the training data. 
mspe_by_sel_seasons

We also tried GAM on Covid-19 vaccine uptake. The training data showed goodness-of fit for the observed uptake: 
covid_train

And the prediction showed reasonable forecast (starting from Sep 1, 2024), with MSPE 0.0012.
covid_test

### DIFF
--- a/scripts/gam_eda.R
+++ b/scripts/gam_eda.R
@@ -6,7 +6,7 @@ library(mgcv)
 ### national flu data ###
 
 ##### data cleaning #####
-nis <- read.csv("../data/nis_flu_national.csv")
+nis <- read.csv("data/nis_flu_national.csv")
 head(nis)
 nis$time_end = as.Date(nis$time_end)
 
@@ -42,7 +42,7 @@ nis_train %>%
   geom_point(aes(x=elapsed,y=estimate)) +
   geom_line(aes(x=elapsed,y=fitted_values,color=season)) +
   theme_bw()
-ggsave("plot/fitted_value_by_season.jpg",units="in",
+ggsave("plot/fitted_value_by_season_lognormal.jpg",units="in",
        width=6,height=4)
 
 
@@ -73,7 +73,7 @@ ggplot(nis_test) +
   geom_point(aes(x=elapsed,y=estimate))+
   geom_line(aes(x=elapsed, y=pred_terms,color=season)) +
   theme_bw()
-ggsave("plot/pred_by_season.jpg",units="in",width=6,height=4)
+ggsave("plot/pred_by_season_lognormal.jpg",units="in",width=6,height=4)
 
 
 ## sequentially add season ##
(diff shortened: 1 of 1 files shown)


swo:
I'm not expert in this method, so take my observations with a grain of salt:

> When fitting the model with national flu vaccine uptake, we found that the fitted line by each season shares the same shape and the intercept differs by season.

It seems like a simple intercept term isn't appropriate here, because it means that, to achieve lower-than-average end-of-season uptake, you need to have negative predictions for start-of-season uptake. Maybe this isn't terrible for a score like MSPE but it does give a very weird smell. Are there ways to put constraints on these functional forms?

> It is expected to see that MSPE decreased as more seasons are included in the training data.

Does this do cross-validation of any kind? If this is including 2009, then 2009 and 2010, then 2009-2011, etc., then one might argue that later seasons are somehow easier to predict than earlier ones, and that's what's driving this signal.

> We also tried GAM on Covid-19 vaccine uptake. The training data showed goodness-of fit for the observed uptake

I guess the question is what kind of performance improvement we get relative to L+L, over what horizons, with what scores, etc.

eschrom:
These are some excellent explanations and illustrations, @Fuhan-Yang! Thanks for such a clear walk-through. Here are some thoughts to add to the discussion:

Regarding season effects:

We definitely want season to be a "random effect." This is what you have already done: each season is constrained by, but deviates somewhat from, the overall average curve. In other words, each season is informed by, but can still differ from, all the others. This is called "partial pooling" of information, and it has several advantages:
- It reflects our actual belief that the fundamental processes underlying uptake are consistent year-over-year, though the details may differ.
- It uses fewer parameters overall and so reduces the risk of overfitting and typically fits faster.
- It provides a sound basis for predicting new seasons. For example, given no information whatsoever about a new season, we can at least guess that it will look like the average of past seasons.

For all these reasons, I recommend you stick with this approach and not bother with a model in which each season is fit totally separately.

Regarding MSPE:

We do _not_ necessarily expect MSPE for a brand new season to decrease as more past seasons are included in the training data. As you train on more past data, the model gets a clearer picture of: a) the overall "average" season, and b) the degree of variability around that "average." But when you predict a new season, only a) really matters - the model has no choice but to expect that the new season will look like the "average" of past seasons. So MSPE is low when the test seasons are similar to the average of the training seasons, and MSPE is high when the test seasons are dissimilar to the average of the training seasons. This does not depend on the relative numbers of training vs. test seasons, but rather on the particular order of the seasons (especially where "outlier" seasons happen to be in that sequence). So your bar plot of MSPE vs. number-of-training-seasons is perfectly plausible.

We _do_ expect MSPE for the end of a season to decrease as more observations from the early part of the season are included in the training data. As the model sees more data about a season, it learns more about the specific ways in which that season differs from the average of all the others before it. But, perhaps counterintuitively, this effect decreases as more full previous seasons are included in the training data. Thus, if you train on all of 2009/10 and part of 2010/11, with the goal of predicting the rest of 2010/11, you should see MSPE decrease noticeably as more of 2010/11 is included in the training data. But if you train on all of 2009/10, 2010/11, ..., 2022/23, and part of 2023/24, with the goal of predicting the rest of 2023/24, you might not see MSPE decrease hardly at all as more of 2023/24 is included in the training data. To top it all off, even this predicted pattern in MSPE may not be particularly strong here, since you are only allowing the Y-intercept to vary by season.

PR #210, opened 2025-09-09
Clarify uv virtual environment

eschrom:
In my last update to the README, I forgot to clarify that there may be different commands to enter a uv virtual environment. This depends on your operating system, which terminal you are using, whether you use custom venv names, etc., so it can differ from user to user. The README now clarifies that `.venv/bin/activate`, `.venv/Scripts/activate`, etc. are just examples of what the enter-my-virtual-environment command might be.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,7 +16,7 @@ Use <https://github.com/CDCgov/nis-py-api> for access to the NIS data.
 
 ## Getting started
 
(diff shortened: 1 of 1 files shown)


Fuhan-Yang (approved):
Thanks for clarifying!!

PR #211, opened 2025-09-15
EDA on number of basis functions given cumulative and incidental uptake data

Fuhan-Yang:
Before switching to Bayesian framework, I did a quick EDA based on today's discussion: how does gam perform given different number of basis functions on cumulative and incidental data?

Conditions: 
1. Flu vaccine uptake data from NIS is used.  
2. Model is assumed to be log-normal distributed (log link function)
3. Cubic basis spline function is used
4. Incidence data is calculated as the difference of cumulative data, and the first "0" point is trimmed to ensure all the data is eligible for log link function.
5. Training and test data is split on 2023-07-01. Season 2023/2024 is used for test. 

Methods:
1. GAM model is fitted given the number of basis functions (k) varied between 5 and 15. 
2. Fitted values and predictions are plotted for each k.
3. MSPE is calculated between test data and prediction for each k. 

Results: 
gam_k_cumu_fitted
As we found out in the morning that more basis functions are needed to better fit the smooth shape of the cumulative data, this also applies to prediction:
gam_k_cumu_pred
k needs to be at least 10 to fit and predict the smooth sigmoid shape of the cumulative data. 

Thus, followed by the thought that incidental data may need fewer basis functions as it includes the turning shape by itself, we've used the same method on incidental data: 
gam_k_inc_fitted
Interestingly, it is found that k is needed to be at least 8 to fit the peak of the incidental vaccine uptake. I speculate that more basis function still provides higher flexibility to model the sharp turning (peak) of the incidence data.

This also applies to prediction: 
gam_k_inc_pred

But the peak can be fitted and predicted when k is at least 8, which is lower than the k for cumulative data. Thus, our assumption is valid that it is easier to capture the wiggliness of the incidence data than cumulative data. 

Also confirmed this using MSPE between test and prediction across k for incidence and cumulative data:
all_mspe_gam_k

We can see that incidence data has pronounced advantage in predicting vaccine uptake. Given the results, I will move forward with trying incidence data on the Bayesian framework, first. 

I also searched if there is basis function that is similar to sigmoid function, but because the basis functions are naturally polynomials, there is no basis function that has sigmoid shape. Increasing k to achieve the sigmoid shape of the smooth function is the common approach, as we expected. This also supports us to move forward with incidence data.

### DIFF
--- a/scripts/gam_k.R
+++ b/scripts/gam_k.R
@@ -1,0 +1,180 @@
+rm(list=ls())
+library(dplyr)
+library(ggplot2)
+library(mgcv)
+
+### national flu data ###
+
+##### data cleaning #####
+nis <- read.csv("data/nis_flu_national.csv")
+head(nis)
+nis$time_end = as.Date(nis$time_end)
+
+nis %>%
+  mutate(time_end=as.Date(time_end),
+         season_start_year= gsub("(\\d{4})/\\d{4}","\\1",season)) %>%
+  mutate(season_start=as.Date(paste0(season_start_year,"-07-01"))) %>%
+  mutate(elapsed=as.numeric(time_end-season_start))%>%
+  mutate(season=factor(season))%>%
+  group_by(season) %>%
+  mutate(inc=c(0,diff(estimate))) %>%
+  filter(inc !=0) %>%
+  select(-season_start_year) -> nis
+
+nis <- nis[order(nis$time_end),]
+
+train_end <- as.Date("2023-07-01")
+
+#### First 11 seasons as training, the rest 4 seasons as testing ###
+nis %>%
+  filter(time_end < train_end) -> nis_train
+
+nis %>%
+  filter(time_end >= train_end) -> nis_test
+
+##### model fitting ######
+# add log-normal link function by manually log-transforming response variable
+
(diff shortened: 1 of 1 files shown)


Fuhan-Yang:
So GAM fits the data in this order:
1. Given the knots and degree of the basis function, create multiple basis functions, like: 
image

2. Confront the data to fit the coefficients for each basis function, the coefficient adjust the amplitude of the basis function, like if we just sample coefficients for each basis function, we can get: 
image

3. So we confront the coefficients with the data, eventually reach the optimal curvature and height of the data: 
image

So 10 knots can't fit 10 data points perfectly because they are not directly used to calibrate to the data. It is the coefficients that fit the data (along with the smoothing parameter for penalty). It's like a hierarchy, the number of knots and the degree of basis function are selected first, then the coefficients will be fitted based on them. In mgcv in R, they did a trick to choose the number of knots when fitting the coefficients (it's called generalized cross validation), but in Bayesian framework, currently we fixed the knots and the degree and only fit the coefficients and the smoothing parameter.

Fuhan-Yang:
@eschrom  This is a great advice! Yes, the MSPE can't be compared between cumulative and incident data. Thanks for the suggestion!

PR #212, opened 2025-09-16
Script for Figures

eschrom:
This is just a script to make figures that can be cobbled together into multi-panel displays. It is probably less important to review the code than to review the figures themselves, which can be found elsewhere.


eschrom:
I am going to merge this without further review, since this is just scratchwork that will not remain in the repo long term.

PR #213, opened 2025-09-19
GAM draft in numpyro

Fuhan-Yang:
- Added a document to describe the math of GAM inference in Bayesian framework
- Added a script to implement the math in the document.

So far, only the main effect (uptake ~s(elapsed)) is considered. Random effect will be added once the inference on main effect is reviewed and improved.

### DIFF
--- a/docs/gam.md
+++ b/docs/gam.md
@@ -1,0 +1,77 @@
+# Overview
+
+GAM (generalized additive model) models the relation between vaccine uptake ($y_i$) and the smooth version ($f(.)$) of elapsed variable (the number of days after vaccine roll-out) ($x_i$) and the random effect introduced by season ($u_j$).
+
+```math
+y_i = f(x_i) + u_j + \beta_0
+
+```
+
+where $f(x_i)=\sum_{k=1}^{K}{\beta_k}{B_k(x_{i,k})}$.
+
+$\beta_0$ is intercept for main effect.
+
+Writing in matrix form, this is:
+
+```math
+
+y = X\beta + Z\u + \beta_0
+```
+
+$y$ is a vector of observed vaccine uptake, $X$ is the design matrix of basis function with $N \times k$ dimension, where $N$ is the number of observations and $k$ is the number of basis functions used. Each element in $X$ is the value of the basis function evaluated at the predictor elapsed ($B_k(x_{i,k})$).$\beta$ is a vector of coefficients that control each basis function. $X\beta$ is the main effect that is the same across all the level in a group (in our case is season).
+
+The loglikelihood function is:
+
+```math
+Loglik(\beta, \lambda |y) = Loglik(y| \beta) - \lambda \beta^TS\beta
+```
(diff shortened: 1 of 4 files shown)


docs/gam.md
```diff
@@ -0,0 +1,92 @@
...
+Writing in matrix form, this is:
+
+```math
+
+g^{-1}(E(y)) = X\beta + Zu + \beta_0
+```
+
+$y$ is a vector of observed vaccine uptake, $X$ is the design matrix of basis function with $N \times k$ dimension, where $N$ is the number of observations and $k$ is the number of basis functions used. Each element in $X$ is the value of the basis function evaluated at the predictor elapsed ($B_k(x_{i})$).$\beta$ is a vector of coefficients that control each basis function. $X\beta$ is the main effect that is the same across all the level in a group (in our case is season). $Z$ is a random-effect design matrix to define the relation between levels in a group. $u$ is a vector representing season-specific intercept. Here, we assume the model follows lognormal distribution and the link function is $log()$.
+
+## Bayesian framework
```
eschrom, line 23: Most of this is beyond my comprehension at the moment - I would need a better fundamental knowledge of the spline / GAM approach to really get the Bayesian version.

scripts/gam_demo.py
```diff
@@ -0,0 +1,172 @@
...
+    S_inv = jnp.linalg.inv(S)
+
+    beta = numpyro.sample(
+        "beta", dist.MultivariateNormal(beta_mu_array, S_inv * sigma / lam)
+    )
+
+    # Identifiability constraints: 1^TX\beta = 0
+    Identity = jnp.array([1.0] * X.shape[0])
+
+    if Identity @ X @ beta.T == 0.0:
```
eschrom, line 148: What does this `@ X @ ...` syntax mean? It's new to me.

eschrom (approved):
Sorry it took me so long to look at this! The numpyro formulation looks fine to me. I don't know if you are still running into trouble, but if so, making sure your priors are very close to realistic values is a good start.

scripts/gam_demo.py
```diff
@@ -0,0 +1,172 @@
...
+
+    Return:
+    np.ndarray
+        the penalty matrix
+    """
+    ## given cubic spline, the penalty matrix is the squared second-order derivatives ##
+    ## Sij = int{B_i''(x) B_j''(x) dx} = int{x_i}{x_j}
+    assert bss.order == 4, "Penalty matrix is only defined for cubic splines"
+
+    derive_order = 2
```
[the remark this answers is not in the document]
Fuhan-Yang: So this is something I'm not sure...there are other options (with different orders in basis function and derivatives) that may perform better than the cubic spline? So I will keep this for now

scripts/gam_demo.py
Fuhan-Yang: @ means matrix multiplication, Identity @ X @ beta.T is $1^{T}X\beta = 0$

PR #214, opened 2025-09-25
Model Tweaks

eschrom:
My first tweak to the Logistic Plus Linear (LPL) model was going to be a mixture model of sorts, in which we add together various sigmoid curves that are assigned different weights. But the more I think about this, the more it seems overly complex and unjustified, since different shapes of sigmoid curves don't really correspond to different interpretable processes of uptake.

So instead, I designed and implemented a Logistic Splice Linear (LSL) model. Whereas LPL adds a logistic and a linear over the x-axis of time, LSL switches from logistic to linear at some specific point in time $x_k$. Because the final curve must be continuous and smooth, the logistic parameters and $x_k$ together determine the linear slope and intercept. Thus, LPL and LSL have the same number of parameters.

Here is the distribution of log MSPE scores for all state x season trajectories, when the model is trained through 2023-09-01 and used to forecast the 2023/2024 season.

LPL:
image

LSL:
image

As you can see, LSL helps the fit to the training data but hurts the fit to the test data. Looks like classic overfitting. LSL has no extra parameters, so it must be an inherently more flexible shape.

This is not the cut-and-dry answer I was looking for (i.e. LSL clearly outperforms the LPL in every regard or vice-versa). In particular, the enhanced flexibility of the LSL might be a liability given our current hierarchical structure, but a strength given some other hierarchical structure.

So, I cannot keep the two or three planned tweaks to our original model as independent, modular endeavors. Instead, I must see how they interact. Next up in this draft PR, I will allow state x season hierarchical effects (not just state and season separate hierarchical effects) for both LPL and LSL. 

@swo @Fuhan-Yang feel free to share alternative thoughts/interpretations if you have them.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,7 +1,7 @@
-RUN_ID = test
+RUN_ID = lsl
 TOKEN_PATH = scripts/socrata_app_token.txt
 TOKEN = $(shell cat $(TOKEN_PATH))
-CONFIG = scripts/config.yaml
+CONFIG = scripts/config_lsl.yaml
 SETTINGS = output/settings/$(RUN_ID)/
 RAW_DATA = output/data/$(RUN_ID)/
 MODEL_FITS = output/fits/$(RUN_ID)/
--- a/iup/models.py
+++ b/iup/models.py
@@ -80,6 +80,377 @@ class UptakeModel(abc.ABC):
     mcmc = None
 
 
+class LSLModel(UptakeModel):
+    """
+    Subclass of UptakeModel for a mixed Logistic Splice Linear model.
+    """
+
+    def __init__(self, seed: int):
+        """
+        Initialize with a seed and the model structure.
+
+        Parameters
+        seed: int
+            The random seed for stochastic elements of the model, to be split for fitting vs. predicting.
+        """
+        self.rng_key = random.key(seed)
+        self.fit_key, self.pred_key = random.split(self.rng_key, 2)
+        self.model = LSLModel._logistic_splice_linear
+
+    @staticmethod
+    def _logistic_splice_linear(
+        elapsed,
+        N_vax=None,
+        N_tot=None,
+        groups=None,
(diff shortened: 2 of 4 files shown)


eschrom:
Before diving into state x season hierarchical effects, I have expanded the set of parameters that have hierarchical effects at all.

For the original LPL model, the logistic asymptote and linear slope had hierarchical effects. Allowing the logistic exponent to have hierarchical effects as well, the log MSPE distribution across all trajectories becomes:
image


For the original LSL model, the logistic asymptote and splice point had hierarchical effects. Allowing the logistic exponent to have hierarchical effects as well, the log MSPE distribution across all trajectories becomes:
image

You can see that LPL performs basically the same, regardless of whether the logistic exponent can vary hierarchically. LSL performs markedly better, particularly in forecasting, when the logistic exponent can vary hierarchically. 

Comparing all 4 plots, it seems that LSL with the extra hierarchical parameter is now the best performing model all-around. One note of danger about LSL, though: when you only have a few data points in a current season and you want to forecast, there is _no information_ available for the model to learn $x_k$ for the current season. The model will have to choose whatever the average value in the past was. This could be part of the explanation for why LSL performs better for training but not testing data.

Next step: try allowing the cross of season x state to be a hierarchical effect, in addition to season and state separately.

eschrom:
The splice point tends to happen about halfway through the season, which is Jan 1ish (since the flu "season" starts on July 1 and ends the following June 30). Not sure if this means that predictions of Jan 1 with very little in-season data are very good. More exploration ahead!

eschrom:
The splice point is indeed fitted with along with the other model parameters. It even varies hierarchically according to state and season.

Meanwhile, the linear slope and intercept for the second half of the curve are not fitted parameters. Instead, they are calculated from the height and slope of the logistic at the splice point, ensuring that the entire curve is both continuous and smooth.

PR #217, opened 2025-12-27
Docs & onboarding

swo:
Changes so that the code runs out of the box, i.e., changes I needed to make to get the code to run in my hands.

- Update the config to deal with updates to NIS data
- Include a check to ensure enough data makes it through pre-processing
- Update Makefile to ensure the dependencies get propagated (i.e., don't use *directories* as targets)
- Update docs to point at mkdocs
- Update project ownership
- Update config to include needed column `sample_size`
- Remove a typo in the model setup

Builds off #216


swo:
Closing in favor of #220

Issue #221, opened 2025-12-29
Remove unneeded dependencies


swo:
Maybe these?

- scikit-learn
- datetime (which is in the standard library)
- pandas?
- pathlib (also in standard library)
- typing-extensions (I don't think we should use typing that isn't in the Python 3.13 standard library)

Some should be moved to dev:

- pre-commit
- ipympl

PR #222, opened 2025-12-29
Update optional dependencies

swo:
I think `scikit-fda` is only needed for the experimental GAM work, but it blocked me from `uv sync` because I was missing some system dependencies.

Rather than document what those system dependencies are, I think a better idea is to move these experimental dependencies into an optional group. (You can install those with `uv sync --all-groups`.)

See also #221

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -24,8 +24,6 @@ dependencies = [
     "pathlib>=1.0.1,<2",
     "streamlit>=1.45.1,<2",
     "scipy>=1.15.3",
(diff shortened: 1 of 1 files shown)


swo:
I removed ipykernel and pre-commit from the main dependencies, since they were already also in `dev`

(To resolve #221 I'll look through all the other dependencies and see which ones are needed, but that will be easier once we have a working pipeline again.)

swo:
Merging based after sync discussion

PR #248, opened 2026-01-08
GAM updates

Fuhan-Yang:
Refactored GAM model: 

1. No sum-to-zero constraint is needed because we only use one smooth term (cubic spline)
2. Added intercept.

So far, given cubic spline function with 10 basis function, ignoring group factors (season), the GAM fit on National flu vaccine uptake looks like: 

image


Fuhan-Yang:
Given 5 basis functions: 
image

PR #251, opened 2026-01-09
Two logistics

swo:
In a quick notebook check, I got a very good fit from just taking the sum of two normal curves. Each season/state combo has 6 parameters (position, scale=SD, and height, for each of the 2 curves).

I started to think about how to actually implement this, and it looks like you have about the same number of parameters as in the LPL model. There's also some pain of refactoring the code to accept multiple models.

### DIFF
--- a/scripts/demo_incident.ipynb
+++ b/scripts/demo_incident.ipynb
@@ -1,0 +1,394 @@
+{
+ "cells": [
+  {
+   "cell_type": "code",
+   "execution_count": 2,
+   "id": "94360df5",
+   "metadata": {},
+   "outputs": [],
+   "source": [
+    "import datetime as dt\n",
+    "\n",
+    "import altair as alt\n",
+    "import numpy as np\n",
+    "import polars as pl\n",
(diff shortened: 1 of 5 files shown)


scripts/demo_incident.ipynb
```diff
@@ -0,0 +1,394 @@
...
+     "execution_count": 4,
+     "metadata": {},
+     "output_type": "execute_result"
+    }
+   ],
+   "source": [
+    "def predict(t: pl.Series, loc1, loc2, amp1, amp2, sig1, sig2) -> pl.Series:\n",
+    "    \"\"\"Predict that incident uptake is the sum of two normals\"\"\"\n",
+    "    y1 = amp1 * scipy.stats.norm(loc=loc1, scale=sig1).cdf(t)\n",
+    "    y2 = amp2 * scipy.stats.norm(loc=loc2, scale=sig2).cdf(t)\n",
```
Fuhan-Yang: Is this still cumulative space?

PR #254, opened 2026-01-13
Demo: Random forest regression

swo:
@Fuhan-Yang @afmagee42 In the spirit of asking what the data can tell us, I did a simple thing, making a random forest regression.

It's kind of hacky: to predict Apr 1 coverage, using data through Jan 1 (say), I train the model on `X` = Aug, Sep, Oct, Nov, Dec, Jan coverages from prior seasons, each as their own feature, and `y` = Apr 1 coverages from those seasons.

Nevertheless, it does pretty well: it makes estimates that are within ~2 p.p. by Jan 1 and ~5 p.p. by Dec.

By looking at feature importances over the forecast horizons, by ~Jan 1, the model is basically just looking at that state's coverage and projecting forward. State is an important feature only for the Aug & Sep projections (i.e., before there is no data). So this indicates yet more strongly to me that the important data for forecasting is what's going on in each state *in the forecast season*, i.e., in the interaction term.

image
where "coverage" is the sum of the importances of the coverage features

@Fuhan-Yang `OrdinalEncoder()` came in handy; that saved me manually coding up the indices. Good find!


afmagee42:
That's pretty convincing to me.

swo:
I made a mistake: I was using ordinal encoding rather than dummy (which scikit calls "one hot") encoding. This means the RF was assuming that 2021/2 was like 2020/1 until it got other information, and similarly tried to make alphabetically-adjacent states similar.

So I threw out season (which maybe I shouldn't have) and encoded geographies as dummy variables. This makes the point even cleaner: in Sep (t=-7), you have about equal information from the state's name and from the Aug & Sep data. But by Oct you're getting 80% of info from Aug/Sep/Oct coverage values. By Jan the prior seasons' uptake info is useless and you just project forward for that state.

image

PR #255, opened 2026-01-16
EDA on model hierarchy

Fuhan-Yang:
I fitted two models in `brms` using all data (ends on 2024-04-30): 1. end-of-season uptake ~ elapsed + 1|season + 1|geography 2. end-of-season uptake ~ elapsed + 1|season + 1|geography + 1|season:geography. While both models fit the data visually well, the model with interaction didn't converge. 

Fitted value from model without interaction and data (similar as model with interaction)
image

Diagnostics of model with interaction: 
image

Note that the model is not optimized, the prior is default non-informative priors and the link function is lognormal, not beta-binomial. So partition of the error (is it from group factor or interaction term) is not optimized.

Because fitting one model in `brms` takes more than 1 hour, I switched to `lmer`, which can fit a linear mixed model as in `brms`, but using restricted likelihood. Here I evaluated the same two models. I tried to evaluate how much interaction terms contribute in forecasting as forecast dates move forward, but I still got failure in converging the model with interaction. From model output, there are 774 parameters for season-state interactions, and we have 6859 data points, I wonder if the failure is because there are way too many parameters for this data size. This is useful to consider the next step, as interaction terms largely scale up the number of parameters, that if we will overspecify the model. 

I also fitted the LPL model in numpyro with data from the beginning of 2009/2010 until 2022-09-01 and calculated the MSPE between the data and forecast after 2022-09-01:
image

Besides the forecast from 2021/2022 is worse than other seasons across all states, and the forecast from Tennessee seems worse across all seasons, we can always observe some single combos of season and state that didn't forecast well. This indicates there must be some kind of deviance haven't been captured yet. Instead of having interaction terms, can we just add an error term for each state-season combo, and estimate the error? @swo @afmagee42

### DIFF
--- a/scripts/lpl_eda.ipynb
+++ b/scripts/lpl_eda.ipynb
@@ -2,7 +2,7 @@
  "cells": [
   {
    "cell_type": "code",
-   "execution_count": 15,
+   "execution_count": null,
    "id": "8b39843d",
    "metadata": {},
    "outputs": [
@@ -18,13 +18,11 @@
     }
    ],
    "source": [
-    "import polars as pl\n",
-    "import numpyro\n",
-    "import jax.numpy as jnp\n",
-    "from numpyro.infer import MCMC\n",
-    "import pickle\n",
     "from datetime import date\n",
+    "\n",
     "import altair as alt\n",
+    "import polars as pl\n",
+    "\n",
     "alt.data_transformers.enable(\"vegafusion\")"
    ]
   },
@@ -36,7 +34,7 @@
    "outputs": [],
    "source": [
     "raw = pl.scan_parquet(\"../data/raw.parquet\")\n",
-    "raw_state = raw.filter(pl.col(\"geography_type\")!=\"nation\").collect()\n",
+    "raw_state = raw.filter(pl.col(\"geography_type\") != \"nation\").collect()\n",
     "raw_state.write_parquet(file=\"../data/raw_state.parquet\")"
    ]
   },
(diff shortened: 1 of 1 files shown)


Fuhan-Yang:
The residual histogram looks like ~N(0, sigma):
image

afmagee42:
Hang on a second, that can't be right. How are there that many thousands of observations? There are 50-ish states and maybe 2 decades of data, there shouldn't be more than like 1000 end of season uptakes.

Fuhan-Yang:
> Hang on a second, that can't be right. How are there that many thousands of observations? There are 50-ish states and maybe 2 decades of data, there shouldn't be more than like 1000 end of season uptakes.

The data includes the number of days from seasonal start (August to next July, 9 data points including normal and leap year) for each season, 15 seasons and 54 states, that give us 9*15*54 = 7290 data points (note the previous model is end_uptake ~ elapsed + 1|season + 1|geography). While the current model only considers season and geography, end_of_season uptake is the same across each level of season and geography. The unique number of data points will be 774, giving the similar histogram: 

image

Fuhan-Yang:
lm_eda
The fitted values from `lm(uptake_on_April1 ~ season + geography)` are plotted as red points with the confidence interval from the data (as error bar). We can see some of the points fell out of the interval, and this is not consistent for a season across all states or for a state across all seasons. There are 36.1% of the total points that have fallen out of the confidence interval (261/723).

PR #261, opened 2026-01-28
Compute index outside of fit() and predict()

swo:
# What's done

- Pass model parameters and MCMC parameters to the model object upon instantiation.
- Do indexing upon model object instantiation.
- Do the indexing by expanding the input data with columns like `season_idx` and `geo_idx`, which avoids a lot of the complicated mapping logic.
- Split the data handed to the model into fitting and prediction data upon instantiation. Each model object now corresponds clearly to a single forecast date.
  - Note that this kind of restricts us to retrospective forecasting: the code now can't be handed one dataset for fitting and a totally separate one for prediction. This is fine, since this project is all about retrospective forecasitng for now.
- Make a new `.model()` method that takes in just the dataframe and pulls out the columns of interest (`elapsed`, `N_vax`, `N_tot`, and the group-index columns), to pass to `_logistic_plus_linear()` (along with some other object-level values that are computed at instantiation)
- Remove the option to have no grouping factors. We always want to group by `"season"`, at least. (#262 would extend this logic to the rest of the codebase.)
- Update `CoverageModel` to match the new structure
- Move "augmentation" (adding season and elapsed time within season) to preprocessing

# What's not done

There's some bug that's causing there to not be predictions for some states, and other predictions to be impossibly narrow. I'll need to dig into that, but I wanted to have some eyes on this in the meantime.

### DIFF
--- a/iup/models.py
+++ b/iup/models.py
@@ -120,16 +120,16 @@ class LPLModel(CoverageModel):
 
     def model(self, data: pl.DataFrame):
         if "N_vax" in data.columns:
-            N_vax = data["N_vax"].to_numpy()
+            N_vax = jnp.array(data["N_vax"])
         else:
             N_vax = None
 
         return self._logistic_plus_linear(
             N_vax=N_vax,
-            elapsed=data["elapsed"].to_numpy(),
+            elapsed=jnp.array(data["elapsed"]),
             # jax runs into a problem if you don't specify this type
             N_tot=jnp.array(data["N_tot"], dtype=jnp.int32),
-            groups=data.select([f"{group}_idx" for group in self.groups]).to_numpy(),
+            groups=jnp.array(data.select([f"{group}_idx" for group in self.groups])),
             n_groups=len(self.groups),
             n_group_levels=self.n_group_levels,
             **self.model_params,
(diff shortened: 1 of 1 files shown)


Fuhan-Yang (commented):
This version makes all the model instances use the same shared raw data with specified `forecast_date` to allow splitting between training and testing data. The raw data for each instance is preprocessed in the same way (including `augment_data`, `augment_scaffold`, value to index mapping from previous version). This seems reasonable to me, as we don't need to differ season start date or `N_tot` between training and testing data, nor across different model instances, plus we want them to share the same index-mapping. This looks good to me!

For the issue of prediction interval, the prediction now is the same as the data for all the sample_ids. Not sure why for now

iup/models.py
```diff
@@ -16,149 +17,148 @@
...
+        self.date_column = date_column
+        self.forecast_date = forecast_date
+        self.groups = groups
+        self.model_params = model_params
+        self.mcmc_params = mcmc_params
+        self.fit_key, self.pred_key = random.split(random.key(seed), 2)
+
+        # input validation
+        assert "season" in self.groups
+        assert {self.date_column, "elapsed", "N_vax", "N_tot"}.issubset(
```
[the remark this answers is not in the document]
swo: "Elapsed" is a function of observation date and season start date, so I think it's actually better to compute it once, rather than multiple times

iup/models.py
```diff
@@ -131,9 +137,9 @@ def model(self, data: pl.DataFrame):
 
     @staticmethod
     def _logistic_plus_linear(
+        N_vax: np.ndarray | None,
```
Fuhan-Yang: I kind of prefer to convert all the variables here into `jnp.ndarray` as this type is the primary use for numpyro (for reason like automatic differentiation and GPU use). The inconsistency of data type for `N_tot`, `N_vax` and `elapsed` also look strange...as they perform similar role in the model

swo: fair; I'll make that change

PR #269, opened 2026-02-06
End-of-season absolute difference by forecast date

Fuhan-Yang:
A draft to visualize how the end-of-season absolute difference (eos abs diff) by forecast date:

1. Summary of eos abs diff for all states:
image

The summary box plot for the eos abs diff for all states is kind of expected: it performs the worst at season start and gets better (notice the drop at September, maybe that indicates the curve reaches plateau?)

2. eos abs diff for individual state:
image

By dissecting how each state contribute to the summary plot, we can tell some states (the first three rows) have good performance that their eos abs diff are pretty low at season start and is pretty consistent across the season. The latter three rows have the similar pattern of performing bad and getting better. @swo

### DIFF
--- a/scripts/plot_data.py
+++ b/scripts/plot_data.py
@@ -93,6 +93,19 @@ def gather_n(df: pl.DataFrame, n: int, col_name="_idx") -> pl.DataFrame:
     )
 
 
+def hightlight_state(df, month, order_n, value="score_value", state_var="geography"):
+    """Return a dic with a list of states (n = order_n) that has lowest score(best) and highest score(worst) at a given month"""
+    sorted_state = (
+        df.filter(pl.col("month") == month)
+        .sort(pl.col(value))
+        .select(state_var)
+        .to_numpy()
+        .ravel()
+        .tolist()
+    )
+    return {"best": sorted_state[:order_n], "worst": sorted_state[-order_n:]}
+
+
 if __name__ == "__main__":
     p = argparse.ArgumentParser()
     p.add_argument("--config", required=True)


swo (changes requested):
- See #271 for my suggestion on how to do the pipeline around the plots
- I'm still confused why you need `eos_abs_diff_by_state.svg` when you have `scores_increasing.svg`. There are showing the same thing?
- I'm similarly confused about `eos_abs_diff_summary.svg`. We have only ~50 states so it's probably better to show how each state's score improves over time, rather than summarize over them, to produce the "cone" of scores in `eos_abs_diff_summary.svg`, which obscures whether, for example, each state has a steadily improving score, or whether they individually bounce around.
- I'm curious about this problem with `.collect()`. It's not a bad practice to use lazy data frames, but the data here aren't that large. In the vignette, the data files are ~1 Mb. This makes me wonder if there's something else going on.

Fuhan-Yang:
> See #271 for my suggestion on how to do the pipeline around the plots

The change looks good to me!

>     I'm still confused why you need `eos_abs_diff_by_state.svg` when you have `scores_increasing.svg`. There are showing the same thing?

`eos_abs_diff_by_state.svg` can better visualize how the score from individual state changes over time, instead of mixing all the states together. 
image
As above, when there are more states than in vignette, it's hard to differ the score from which state is decreasing, or staying stable across time. 

>     * I'm similarly confused about `eos_abs_diff_summary.svg`. We have only ~50 states so it's probably better to show how each state's score improves over time, rather than summarize over them, to produce the "cone" of scores in `eos_abs_diff_summary.svg`, which obscures whether, for example, each state has a steadily improving score, or whether they individually bounce around.

Sounds fair to me.

>     * I'm curious about this problem with `.collect()`. It's not a bad practice to use lazy data frames, but the data here aren't that large. In the vignette, the data files are ~1 Mb. This makes me wonder if there's something else going on.

If running all the states, the file by each forecast date is ~170 Mb. It's unavoidably huge when combining files from all forecast dates together (10 files).

swo (commented):
Returning review gated on some changes to figures.

Ping me again when it's time!

PR #281, opened 2026-03-16
Use arrays

swo:
I was thinking about why the data weren't that big on disk but took up lots of space in memory. My guess is that it's because we had data in long format, with many repeated values. On disk, that would be compressed.

So an alternative would be a wide format, but one where we don't put the individual samples in their own column, but instead have a single array column.

### DIFF
--- a/iup/__init__.py
+++ b/iup/__init__.py
@@ -155,6 +155,4 @@ class SampleForecast(Data):
     """
 
     def validate(self):
-        self.assert_in_schema(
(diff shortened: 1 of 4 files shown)


Fuhan-Yang (commented):
The change can run the prediction for all the states and all the seasons! While there seems to be a bug in the prediction...

When running config_vignette.yaml, the `estimate` in the data to plot is ~100 times smaller than the raw data, while the `pred_estimate`, `pred_lci` and `pred_uci` are ~100 times larger: 
Image

PR #290, opened 2026-04-13
Proof of concept for a dagsterized approach

jkislin:
This example has dummy assets as a proof of concept; this is not intended to be merged as written, but rather to demonstrate capability and ease of adoption.

@swo @Fuhan-Yang feel free to ignore if not helpful - I know dagster isn't your current priority but figured I'd share this as we start extending to other projects. It was just a few lines of code to tailor existing work (such as what we have in the RTM branch repos) to your repo. 

Here's an example asset graph based on these additions:

image  

Note that I'm guessing at what the DAG should look like. A short conversation between us and a small amount of development time should be enough to move this from POC to MVP with feature parity with the `Makefile`.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,49 @@
+FROM ubuntu:24.04
+
+RUN apt update -y && apt install git -y
+
+# Python from https://docs.astral.sh/uv/guides/integration/docker/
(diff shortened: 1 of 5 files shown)


swo:
This was useful to see, but I've concluded this isn't the right place for a first prototype.

PR #293, opened 2026-04-14
Add RFModel

Fuhan-Yang:
1. Deleted adding `t` in `scripts/preprocess.py` and added it in separate way in LPLModel and RFModel, because `t` exists for both models but defined differently. Recovered dates relevant to seasons back to `model. params`. 
2. Addded RFModel. 
- `_preprocess()`: shape the data into columns with monthly estimates along with `season` and `geography`. 
- `fit()`: Encode data. Iteratively fitted the monthly estimates that are after the month of `forecast_date` in the seasons before the target season. 
- `predict()`: Predict the monthly estimates that are after the month of `forecast_date` in the target season. 
3. Updated `eval.py`.
4. Updated relevant tests. 

The fitting and prediction of RFModel are set up this way to make sure RFModel generates the same format of predictions as LPLModel: the prediction median and 95% interval by month after forecast_date. But note the uncertainty is different. The prediction interval of LPLModel is the posterior prediction interval, representing the parameter uncertainty and can project to the unobserved data with the same level of uncertainty. While RFModel generates the forecasts by mapping the monthly estimate before forecast date and future monthly estimates, the uncertainty is the bootstrapping uncertainty that can differ by data points. In this case, it will be better to plot the prediction interval of LPLModel as a cone, while as data-wise median and error bar for RFModel. 

The `plot_preds.py` and `plot_scores.py` need rework. In LPLModel, the in-sample prediction and scores are plotted together in a plot, but the current RFModel only produces out-of-sample score and prediction. Need to discuss. Also, current config and scripts only allow to run one model one time. Need to discuss how to restructure the code base to run the models together.


iup/eval.py
```diff
@@ -42,6 +42,7 @@ def mspe(
         .agg(pl.col("score_value").mean())
         .with_columns(score_fun=pl.lit("mspe"))
         .select(grouping_factors + SCORE_COLS)
+        .drop_nulls()
```
swo, line 45: This always makes me nervous. Why do we have nulls here?

iup/eval.py
```diff
@@ -80,5 +81,6 @@ def eos_abs_diff(
             score_fun=pl.lit("eos_abs_diff"),
         )
         .select(grouping_factors + SCORE_COLS)
+        .drop_nulls()
```
swo, line 84: Same q

iup/models.py
```diff
@@ -31,6 +34,10 @@ def __init__(
     ):
         pass
 
+    @abc.abstractmethod
+    def _preprocess(cls) -> pl.DataFrame:
+        pass
```
swo, line 39: There's no need to have a private method in the ABC. The things we want in the ABC are the ones that get called from the outside, so init, fit, predict. We don't care, from the outside, what other private methods different classes have.

iup/models.py
```diff
@@ -98,18 +107,24 @@ def __init__(
         self.mcmc = None
 
     @classmethod
-    def _preprocess(cls, data: pl.DataFrame) -> pl.DataFrame:
+    def _preprocess(cls, data: pl.DataFrame, date_column, params) -> pl.DataFrame:
```
swo: If the preprocessing is common to all models, we should put it in the upstream script. If it's specific to different models, then it shouldn't be in the ABC

iup/utils.py
```diff
@@ -28,3 +31,11 @@ def date_to_season(
 
     year2 = year1 + 1
     return pl.format("{}/{}", year1, year2)
+
+
+def month_order(season_start_month: int) -> List[str]:
```
swo, line 36: If this is only used by `RFModel`, then it should be a private method in that class

iup/models.py
```diff
@@ -313,3 +357,213 @@ def predict(self) -> pl.DataFrame:
...
+        forecast_t = (
+            self.months.index(self.forecast_date.strftime("%b")) - self.end_month_index
+        )
+
+        end_date = datetime.date(
+            self.params["end_year"], self.params["end_month"], self.params["end_day"]
+        )
+
+        # this is true only when target_season is the last season in the data, which is our case for now
+        data_fit = self.data.filter(pl.col("season") != target_season)
```
swo, line 436: Let's put some kind of assertion or something in here

iup/models.py
```diff
@@ -313,3 +357,213 @@ def predict(self) -> pl.DataFrame:
...
+
+        self.models = {}
+
+        ## fit all the data after forecast_t ##
+        for target_t in range(forecast_t + 1, 1):
+            features = ["season", "geography"] + [
+                f"t={t}"
+                for t in range(
+                    1 - self.months.index(end_date.strftime("%b")), forecast_t + 1
```
swo, line 445: This seems a little crazy. Can you keep track of which months you have elsewhere in the model?

iup/models.py
```diff
@@ -313,3 +357,213 @@ def predict(self) -> pl.DataFrame:
...
+                )
+            ]
+
+            # always fit the last month of the season given the months before forecast_date
+            X_fit = self.enc.encode(data_fit.select(features))
+            y_fit = data_fit.select(f"t={target_t}").to_series().to_numpy()
```
swo, line 451: It occurred to me that the random forest regressor can predict multiple features, so we could ask it to forecast coverages in all the remaining dates in the season, not just the last one

iup/models.py
```diff
@@ -313,3 +357,213 @@ def predict(self) -> pl.DataFrame:
...
+
+        return self
+
+    def predict(self) -> pl.DataFrame:
+        data_pred = self.data.filter(
+            pl.col("season")
+            == date_to_season(
+                pl.lit(self.forecast_date),
+                season_start_month=self.params["start_month"],
+                season_start_day=self.params["start_day"],
```
swo, line 465: This is repeat above; this means there should be a `self.forecast_season = ...` in the `__init__()`

iup/models.py
```diff
@@ -313,3 +357,213 @@ def predict(self) -> pl.DataFrame:
...
+            self.params["end_year"], self.params["end_month"], self.params["end_day"]
+        )
+
+        # this is true only when target_season is the last season in the data, which is our case for now
+        data_fit = self.data.filter(pl.col("season") != target_season)
+
+        self.models = {}
+
+        ## fit all the data after forecast_t ##
+        for target_t in range(forecast_t + 1, 1):
```
swo, line 441: Try and refactor this as `self.models = {target_t: some_function(target_t) for ...}`, rather than doing a loop

iup/models.py
```diff
@@ -313,3 +357,213 @@ def predict(self) -> pl.DataFrame:
...
+                    ),  # convert back to month index
+                    target_year=pl.col("season").str.extract(r"^(\d{4})/\d{4}"),
+                )
+                .with_columns(
+                    season_start_date=pl.date(
+                        pl.col("target_year"),
+                        self.params["start_month"],
+                        self.params["start_day"],
+                    ),
+                    target_index=pl.col("target_index").cast(pl.String) + "mo",
```
swo, line 519: this is a good place for `pl.format`

iup/models.py
```diff
@@ -313,3 +357,213 @@ def predict(self) -> pl.DataFrame:
...
+            pl.col("season")
+            == date_to_season(
+                pl.lit(self.forecast_date),
+                season_start_month=self.params["start_month"],
+                season_start_day=self.params["start_day"],
+            )
+        )
+
+        forecast_t = (
+            self.months.index(self.forecast_date.strftime("%b")) - self.end_month_index
```
swo, line 470: OK, this is all too crazy. In `__init__()` and `_preprocess()`, you should take the data from its long, date-valued format into the wide, `t=`-valued format. `fit()` and `predict()` should mostly work with the wide, indexed data. Then, only once, at the end of `predict()`, you'll convert back to long, date-valued data. It's too crazy to be swapping between dates and `t=` indices while you're doing all these computations

iup/models.py
```diff
@@ -313,3 +357,213 @@ def predict(self) -> pl.DataFrame:
...
+            )
+        ]
+
+        X_pred = self.enc.encode(data_pred.select(features))
+        quantile_cols = [f"q={k}" for k in self.quantiles]
+        index_cols = ["season", "geography"]
+
+        preds = []
+
+        for target_t, model in self.models.items():
```
swo, line 490: Another loop to replace with a function

iup/models.py
swo: This would mean each `RFModel` would have only one regression inside it, rather than one per target date

scripts/config_vignette.yaml
swo, line 1: Let's just make a separate config for the RF for now -- you should be able to still run the code using the vignette that calls the LPL model, or one that calls the RF model

Makefile
swo, line 1: let's revert these changes?

swo (changes requested):

iup/eval.py
Fuhan-Yang: This is because when `join` with `data` under `mspe()` in `iup.eval`, the seasons used to fit RF model is not included in the prediction in RF model. This bug was not triggered in LPL model because the predictions in LPL model includes in-sample prediction and out-of-sample prediction, splitted by `forecast_date`. Maybe a better way is to filter to the target season only, we only interested in forecasting performance in target season.

Fuhan-Yang: Per discussion, use mspe for in-sample validation

iup/utils.py
Fuhan-Yang: Updated. Also changed for `date_to_season`

iup/models.py
Fuhan-Yang: Fixed!

iup/models.py
Fuhan-Yang: This has been updated to predicting a vector

Fuhan-Yang: For the crazy indexing...the index has been recovered in the `target_index` in `pred()`. This follows the demo `rf.py`, where the months are indexed relative to the end-of-season month (which is fixed to 0)

iup/models.py
Fuhan-Yang: Updated!

iup/models.py
Fuhan-Yang: This has been updated as now both in-sample and out-of-sample predictions are generated

iup/models.py
Fuhan-Yang: See updated as now it's predicting a vector

Makefile
Fuhan-Yang: done

Fuhan-Yang:
Most of the comments are fixed...except the `null` in `eval.py` and the crazy indexing among `preprocess`, `fit` and `predict`...also the pytest needs update as the prediction target changes. Thanks Scott for taking on this!

PR #334, opened 2026-06-23
Make sure forecast dates are within the date range from the data

### LINKED ISSUE #333: Check the end_year in "season" aligns with the season in "forecast_dates"
In config, when `end_year` under `season` is earlier than the second year in `forecast_dates`, RFModel will create unexpected results, such as wrong forecast dates in forecast.svg or multiple forecast seasons. The solution can be to assert the `forercast_date` not later than the `end_year` in config

Fuhan-Yang:
Add conditionals in Makefile and `get_forecast_dates.py` to make sure the input `forecast_dates` falls into the range of data (indicated by arguments in `season` in config). Resolves #333

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,7 +19,7 @@ FORECAST_DATES = $(shell python scripts/get_forecast_dates.py --config=$(CONFIG)
 PREDS = $(foreach date,$(FORECAST_DATES),$(PRED_DIR)/forecast_date=$(date)/part-0.parquet)
 FITS = $(foreach date,$(FORECAST_DATES),$(OUTPUT_DIR)/fits/fit_$(date).pkl)
 
-ifeq ($(FORECAST_DATES),forecast date out of data range. )
+ifeq ($(FORECAST_DATES),forecast date out of data range.)
 $(error forecast date out of data range, check config)
 endif
 
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-CONFIG = scripts/config_vignette.yaml
+CONFIG = scripts/config_full.yaml
 RAW_DATA = data/raw.parquet
 
 RUN_ID = $(shell python scripts/get_run_id.py --config=$(CONFIG))
@@ -18,6 +18,10 @@ PLOT_SCORES = $(OUTPUT_DIR)/plots/.scores.checkpoint
 FORECAST_DATES = $(shell python scripts/get_forecast_dates.py --config=$(CONFIG))
 PREDS = $(foreach date,$(FORECAST_DATES),$(PRED_DIR)/forecast_date=$(date)/part-0.parquet)
 FITS = $(foreach date,$(FORECAST_DATES),$(OUTPUT_DIR)/fits/fit_$(date).pkl)
+
+ifeq ($(FORECAST_DATES),forecast date out of data range. )
+$(error forecast date out of data range, check config)
+endif
 
 # This variable because the pattern `forecast_date=2020-01-01` confuses make.
 # It thinks `=%` is variable assignment, not pattern matching.
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-CONFIG = scripts/config_full.yaml
+CONFIG = scripts/config_vignette.yaml
 RAW_DATA = data/raw.parquet
 
 RUN_ID = $(shell python scripts/get_run_id.py --config=$(CONFIG))
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -19,6 +19,7 @@ dependencies = [
     "numpyro>=0.16.1,<0.17",
     "jax>=0.5.0,<0.6",
     "scipy>=1.15.3",
+    "pre-commit>=4.6.0",
 ]
 
 [build-system]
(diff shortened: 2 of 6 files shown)


pyproject.toml
```diff
@@ -19,6 +19,7 @@ dependencies = [
     "numpyro>=0.16.1,<0.17",
     "jax>=0.5.0,<0.6",
     "scipy>=1.15.3",
+    "pre-commit>=4.6.0",
```
swo: this should go in dev dependencies

Makefile
```diff
@@ -1,4 +1,4 @@
-CONFIG = scripts/config_vignette.yaml
+CONFIG = scripts/config_full.yaml
```
swo: we should have the vignette be the default

scripts/get_forecast_dates.py
```diff
@@ -18,4 +19,17 @@
...
+        config["season"]["start_month"],
+        config["season"]["start_day"],
+    ):
+        print("forecast date out of data range. ")
+    elif forecast_dates[-1] > date(
+        config["season"]["end_year"],
+        config["season"]["end_month"],
+        config["season"]["end_day"],
+    ):
+        print("forecast date out of data range. ")
```
swo: I hate having to have this magic string at all, but given that we need it, let's only write it once? You'll want an `or` statement

Makefile
```diff
@@ -19,6 +19,10 @@ FORECAST_DATES = $(shell python scripts/get_forecast_dates.py --config=$(CONFIG)
 PREDS = $(foreach date,$(FORECAST_DATES),$(PRED_DIR)/forecast_date=$(date)/part-0.parquet)
 FITS = $(foreach date,$(FORECAST_DATES),$(OUTPUT_DIR)/fits/fit_$(date).pkl)
 
+ifeq ($(FORECAST_DATES),forecast date out of data range. )
```
swo: So minor, but do we need that space at the end of the string??

swo (changes requested):

swo:
I'm also still unclear on whether this was the cause of the issue described in #333

Makefile
Fuhan-Yang: No we don't, just to match the printed output from `get_forecast_dates.py`.

Fuhan-Yang:
Per discussion, the issue #333 is caused by the input data is out of the range of available data. Check has been added.

scripts/config_vignette.yaml
```diff
@@ -41,7 +41,7 @@ alpha: 0.05
 forecast_dates:
   # run forecasts starting Oct, Nov, Dec
   start: 2021-10-01
-  end: 2021-12-01
+  end: 2022-03-01
```
swo: Again, we should keep the *vignette* small, and only make the *full* config larger

scripts/get_forecast_dates.py
```diff
@@ -18,4 +19,15 @@
         eager=True,
     )
 
-    print(*forecast_dates, sep=" ")
+    if forecast_dates[0] < date(
```
swo: This is very stylistic and minor, but I think this would be easier to read as

```python
first_season_date = date(config["season"]["start_year"], ...)
if (forecast_dates[0] < first_season_date) or (forecast_dates[-1] > last_season_date):
    pass
```

the nesting is sometimes hard on the eyes

scripts/preprocess.py
```diff
@@ -94,6 +95,26 @@ def geo_filter(df: pl.DataFrame) -> pl.DataFrame:
     assert isinstance(config, dict)
     geographies = config.get("geographies", None)
 
+    data_start = raw_data["time_end"].min()
+    data_end = raw_data["time_end"].max()
```
swo, line 99: Does it matter that this date could differ by state? Maybe that's a separate issue. But I think it's been a recurrent problem for our visualizations, that some states have different data ranges for different seasons. It might be something we want to have a more robust solution for?

scripts/preprocess.py
```diff
@@ -94,6 +95,26 @@ def geo_filter(df: pl.DataFrame) -> pl.DataFrame:
...
+
+    # ensure data from config is within the available data range #
+    assert (
```
swo: same note as above, about nesting. not critical but just something that makes it harder to follow as a reviewer

swo (changes requested):
The config change is gating; everything else is stylistic or things for your consideration

scripts/preprocess.py
Fuhan-Yang: This check is to provide the inclusive check that the input from config should be within the broadest date range among the states, to avoid the issue that there is no valid month encoding for the dates that are not available in the data (`self.date_crosswalk` in `RFModel`). 

For the states that have different date ranges for different seasons, it leads to varied forecast performance depending on the amount of data used for training. This could be a section included in model evaluation. I will include this showing the figures

scripts/get_forecast_dates.py
Fuhan-Yang: Makes sense, fixed!

swo (approved):

PR #335, opened 2026-06-24
Suggestion: date checks

swo:
If we're doing date checks, I would be a little more robust: we want to make sure these four dates are in order: start of available data, first forecast date, last forecast date, end of data

The thing about returning a string, that make then catches, is very finicky. I'd rather either (1) keep that kind of error handling manual or (2) migrate to Snakemake or some other framework.

Also, again, I think it's important that we use the vignette config by default. If you want to run another config, you should `make CONFIG=scripts/config_full.yaml`.

Finally, I this doesn't throw any error for me? Like, maybe it's nice to have this check, but I'm not getting an error?

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -19,10 +19,6 @@ FORECAST_DATES = $(shell python scripts/get_forecast_dates.py --config=$(CONFIG)
 PREDS = $(foreach date,$(FORECAST_DATES),$(PRED_DIR)/forecast_date=$(date)/part-0.parquet)
 FITS = $(foreach date,$(FORECAST_DATES),$(OUTPUT_DIR)/fits/fit_$(date).pkl)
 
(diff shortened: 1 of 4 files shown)


Fuhan-Yang (commented):
Adding the check in `get_forecast_dates.py` will not stop the subsequent code from running, that's why I explored a bit to add it in Makefile. But yes, it is not safe to rely the check on manually print-out content...

This is not an explicit error but wrong figure output. It is caused by `["season"}["end_year"}` or `forecast_dates` in config being out of range of raw data. I've updated in issue #333 Otherwise, code runs well!

swo:
I saw that too: this is another limitation of `make`, that only errors thrown in the recipes (i.e., the actions in the rules) will stop `make`.

PR #336, opened 2026-07-01
Remove unused arguments & update the forecast dates in config

Fuhan-Yang:


scripts/config_full.yaml
```diff
@@ -38,12 +38,10 @@ models:
 
 forecast_dates:
   start: 2021-07-01
-  end: 2022-04-01
+  end: 2022-03-01
```
swo, line 41: Did you also want to change this date?

swo (approved):

scripts/config_full.yaml
Fuhan-Yang: Yes, let me rename the PR

PR #337, opened 2026-07-01
Add in-sample MSPE

Fuhan-Yang:
The report mentioned using the performance of the fitted models on training data to evaluate their performance. MSPE score is added to evaluate the performance of in-sample prediction.

### DIFF
--- a/scripts/config_full.yaml
+++ b/scripts/config_full.yaml
@@ -38,7 +38,7 @@ models:
 
 forecast_dates:
   start: 2021-07-01
-  end: 2022-04-01
+  end: 2022-03-01
   interval: 1mo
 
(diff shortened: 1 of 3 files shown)


swo:
My recollection is that we jettisoned this concept of "model fit" because the random forest doesn't really work this way.

We were concerned about fit with the parametric, Bayesian curve fits, like the LPL. If we had multiple parametric curve fits, it would make sense to compare their retrospective fits against one another. But it doesn't really make sense to ask about the retrospective fit of LPL vs. RF. RF will always win, and that's not dispositive of anything.

Fuhan-Yang:
Per discussion, remove MSPE in the codebase.

